PR #6, opened 2025-08-06
Add swagger functions to utilities

henrikhenrikpersson:
Moves the swagger middleware functions that were probably used by the work-order service into `libs/utilities`.


services/property/src/middleware/parse-request.ts
```diff
@@ -1,18 +1,11 @@
...
-    parsedQuery: Q extends z.ZodType ? z.infer<Q> : never
-  }
-}
-
 export function parseRequest<T extends z.ZodType, Q extends z.ZodType>(params: {
   body?: T
   query?: Q
 }) {
-  return function (ctx: ContextWithParsedRequestBody<T, Q>, next: Next) {
+  return function (ctx: ParameterizedContext, next: Next) {
```
momentiris, line 8: This change to the types should mean that we no longer get inference from using the middleware. Have you tried it?

henrikhenrikpersson: Yes and you are correct, I am working on it :)

momentiris:
Zod v4 docs uses `import * as z from 'zod'`. Do  you know if it matters if we still use `import {z} from 'zod'`?

henrikhenrikpersson:
> Zod v4 docs uses `import * as z from 'zod'`. Do you know if it matters if we still use `import {z} from 'zod'`?

I don't think so, it looks like you can import it however you want.
```ts
import z from 'zod'
import { z } from 'zod'
import * as z from 'zod'
```

momentiris (approved):

momentiris:
https://github.com/Bostads-AB-Mimer/onecore/pull/6/commits/a6e71e3cc3909edfe314836db00303122bf19521

This commit makes sure we use exported utilities fns for both core + property (who currently converts zod scheman to json schemas)
+ a lot of scattered clean ups/fixes for this the above

momentiris:
https://github.com/Bostads-AB-Mimer/onecore/actions/runs/18716504837/job/53377263020?pr=6

Somehow core tests failing due to missing export from utilities. @svjson any idea?

PR #11, opened 2025-08-12
chore: add turborepo

momentiris:
Adds [turborepo](https://turborepo.com/docs/getting-started/add-to-existing-repository) with basic configuration.

As a starting point, turbo terminal UI lets us run multiple packages in parallell for local development.
I've added some documentation on this in the README, and also linked to turbo docs where more information can be found.

Turborepo can also help us build our packages (with its dependencies to other packages, etc) in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,4 @@ lib-cov
 *.njsproj
 *.sln
 *.sw?
+.turbo
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@
 
 Monorepo containing all base services and applications of the ONECore platform.
 
-
 ## Overview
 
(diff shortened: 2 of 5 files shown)


momentiris:
> LGTM
> 
> 
> 
> Except - why pin npm at an old version (9.9.3) ?

Turbo needs it for "optimizations". I just used the output of my npm --version. But yeah we could use ^10. Not sure if that implies a higher min node version.

svjson:
> But yeah we could use ^10. Not sure if that implies a higher min node version.

It does not. I'm on npm 10.8.2.

svjson (approved):

PR #15, opened 2025-08-14
MIM-563: Incorrect message timestamps

simonkropp:
Dates returned from workorder api seemed to return a local date instead of UTC (message dates are stored without timezone in odoo)

The timestamp for messages were therefore displayed incorrectly in mimer.nu.

Now it localizes the UTC to Swedish time correctly.


johanneskarlsson (approved):
Great!

PR #22, opened 2025-08-25
MIM-547: Create maintenance request on parking space bugfix - some parking spaces were not returned

simonkropp:
seems like we need to join on keycmobj like the original query in order to return all parking spaces.. otherwise only a few are returned.

### DIFF
--- a/services/property/prisma/schema/PropertyStructure_babuf.prisma
+++ b/services/property/prisma/schema/PropertyStructure_babuf.prisma
@@ -64,7 +64,7 @@ model PropertyStructure {
   toDate                   DateTime @map("tdate") @db.DateTime
   timestamp                String   @map("timestamp") @db.Char(10)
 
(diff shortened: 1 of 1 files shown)


services/property/prisma/schema/PropertyStructure_babuf.prisma
```diff
@@ -64,7 +64,7 @@ model PropertyStructure {
   toDate                   DateTime @map("tdate") @db.DateTime
   timestamp                String   @map("timestamp") @db.Char(10)
 
-  parkingSpace    ParkingSpace?    @relation(fields: [parkingSpaceId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  parkingSpace    ParkingSpace?    @relation(fields: [propertyObjectId], references: [propertyObjectId], onDelete: NoAction, onUpdate: NoAction, map: "fkbabufobjbps")
```
johanneskarlsson, line 67: "fkbabufobjbps" - what's that? 😅

PR #23, opened 2025-08-26
Fix: UTH-236 XPand-DB Parking Space Query Gets the Wrong VacantFrom Date

mittistormen:
Parking spaces doesn't get the correct VacantFrom date from the xpand-db query to get parking spaces

Fix:
* A sub query to pick the latest sistadeb date of the contracts of a parking space
* Logic to add take sistadb and add a day to set VacantFrom date

Skärmavbild 2025-08-26 kl  14 16 10

### DIFF
--- a/services/leasing/src/services/lease-service/adapters/xpand/rental-object-adapter.ts
+++ b/services/leasing/src/services/lease-service/adapters/xpand/rental-object-adapter.ts
@@ -313,9 +313,12 @@ const getAllVacantParkingSpaces = async (): Promise<
         )
       })
       //exclude parking spaces with active contracts
-      .whereNull('ac.keycmobj')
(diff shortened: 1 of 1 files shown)


services/leasing/src/services/lease-service/adapters/xpand/rental-object-adapter.ts
```diff
@@ -261,21 +300,26 @@ const getAllVacantParkingSpaces = async (): Promise<
...
       .where(function () {
         this.whereNull('rb.keycmobj').orWhere(
           'rb.blockenddate',
           '<=',
           xpandDb.fn.now()
         )
       })
-      .whereNull('ac.keycmobj')
+      //exclude parking spaces with active contracts
+      .where(function () {
```
mittistormen, line 316: @lalmqvist här!

PR #25, opened 2025-09-01
Fix: No Priority For Applicants

mittistormen:
Since replacing the functionaty for getting rental objects from SOAP to Xpand-DB and changing the listing model the residantial area code is no longer in district code.

* The priority service is now using listing.rentalObject.residentialAreaCode

### DIFF
--- a/services/leasing/src/services/lease-service/routes/listings.ts
+++ b/services/leasing/src/services/lease-service/routes/listings.ts
@@ -702,7 +702,6 @@ export const routes = (router: KoaRouter) => {
         }
       }
 
-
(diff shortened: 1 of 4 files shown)


services/leasing/src/services/lease-service/routes/listings.ts
```diff
@@ -702,11 +703,31 @@ export const routes = (router: KoaRouter) => {
       }
 
       //TODO: get rental object from new xpand-adapter that gets rental objects from db
```
momentiris, line 705: Maybe remove the todo if this is now implemented? 🤔

PR #29, opened 2025-09-03
Infra/build config harmonization step 1

svjson:
### WHAT
- Harmonization of build and test configs
- Common tsconfig.base.json & tsconfig.react.base.json in monorepo root. Modules override what they need, nothing else
- Run all tests from root
- Auto-start docker compose database and create test db on npm run test in leasing - no more manual steps
- Disabled borked test quite in property, because it was abandoned to die alone in the cold under a bridge before electricity was discovered
- Example of splitting test and build tsconfigs in internal-portal-backend
- Some random fixup of jest configs
- Made that same duplicated error-handler.test.ts STFU in test runs to avoid that constant false signalling that something went wrong when it actually did not.

### WHY
- To facilitiate building and testing the repository in one fell swoop
- To avoid special cases and requirements for various modules in github actions
- Inner peace


mittistormen (approved):
Det ser fint ut! Vilken härlig bonus att få testdatabasen uppsatt automatiskt 🔥

PR #35, opened 2025-09-08
Feat/uth 236 get all parking space listings from one core

mittistormen:
Added missing functionality to filter result based on a ContactCode. Scored/internal Listings where the user doesn't have a housing contract is filtered out.


core/src/services/lease-service/helpers/lease.ts
```diff
@@ -0,0 +1,20 @@
+import { Tenant } from 'libs/types/dist'
+
+export const isTenantAllowedToRentAParkingSpaceInThisResidentialArea = (
```
lalmqvist, line 3: Nice 👏

PR #41, opened 2025-09-09
MIM-608: Fetch building on property code in core

simonkropp:
Allows fetching buildings by providing a property code

### DIFF
--- a/core/src/adapters/property-base-adapter/index.ts
+++ b/core/src/adapters/property-base-adapter/index.ts
@@ -423,6 +423,27 @@ export async function getFacilityByRentalId(
   }
 }
 
+type GetBuildingsResponse = components['schemas']['Building'][]
+
+export async function getBuildings(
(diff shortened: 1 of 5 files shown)


core/src/adapters/property-base-adapter/index.ts
```diff
@@ -423,6 +423,27 @@ export async function getFacilityByRentalId(
   }
 }
 
+type GetBuildingsResponse = components['schemas']['Building'][]
+
+export async function getBuildings(
```
johanneskarlsson, line 428: `getBuildingsByPropertyCode` maybe?

johanneskarlsson (approved):
Snyggt! En minor 👏

core/src/adapters/property-base-adapter/index.ts
simonkropp: good point, that is better :)

PR #48, opened 2025-09-16
Feat:UTH-237 Route to Get Note of interests / Applicants should also …

mittistormen:
…get Rental Object for the listing

Rental object is needed as well now that we dont save rental object info to db


lalmqvist (approved):
Ser fint ut! 👏

PR #49, opened 2025-09-16
Feat/UTH-258 mark listing as assigned when lease has been created

mittistormen:
* removed unused code (UTH-259)
* use listing from leasing instead of parking space from XPand to check if parking space is available for leaseing
* set listing to assigned when lease has been created for NON_SCORED listings


core/src/processes/parkingspaces/external/index.ts
```diff
@@ -46,9 +47,11 @@ export const createLeaseForExternalParkingSpace = async (
   ]
 
   try {
-    const parkingSpace = await getParkingSpace(parkingSpaceId)
+    // const parkingSpace = await getParkingSpace(parkingSpaceId)
```
lalmqvist, line 50: 💀 ?

lalmqvist (approved):
Looks good! And nice with all the deleted code 💯

core/src/processes/parkingspaces/external/index.ts
mittistormen: hehe, ja! fixar det 👍

PR #62, opened 2025-09-29
Fix/uth 276 show vat for non scored parking spaces

mittistormen:
NON_SCORED parking spaces is rented out with VAT if the customer doesn't have a contract in the same area as the parking space. VAT needs to be shown in the internal portal.

* VAT shown for NON_SCORED parking spaces in the parking space list.
* VAT shown för NON_SCORED parking spaces on the parking space view.
* fixed column sizet to make it easier to read
* applicants and assign date is now hidden för NON_SCORED parking spaces as well


momentiris (commented):
Code LGTM except for VAT calculation. I would use something like currency.js to handle money calculations in a more robust way, as javascript is known for floating point precision issues.

PR #68, opened 2025-10-06
Feat:UTH-281 Handle Blocks Correctly For Vacant Parking Spaces

mittistormen:
A rental object can have multiple blocks and the vacant parking spaces query should consider the last one of them when deciding if a rental object should be returned or not

* Now using the same functionality for getting blocks of rental object as we already did in the other rental object queries


svjson (approved):
LGTM, så långt som jag kan bedöma

PR #505, opened 2026-04-13
mim-1673: Add tenant information to inspection form

simonkropp:
Show tenant information in inspection form (+ mobile).. fetch all lease data required to replace all the hard coded strings etc.


apps/property-tree/src/features/inspections/ui/InspectionConductDialog.tsx
```diff
@@ -3,14 +3,37 @@ import { useQuery, useQueryClient } from '@tanstack/react-query'
...
 
+/**
+ * Build a TenantInfoCardData view-model from the lease that matches the
+ * inspection. The lease carries everything the card needs (tenant identity,
+ * move-in/out dates, notice status), so no extra tenant lookup is required.
+ */
+const buildTenantInfo = (
+  lease: Lease | undefined
+): TenantInfoCardData | undefined => {
+  const tenant = lease?.tenants?.[0]
```
Crewiz: Pick by leaseContactType === 'INNEHAVARE' explicitly, see isLeaseHolderRelation in services/leasing/.../tenant-lease-adapter.ts:1136-1137]. Otherwise the new card will occasionally show the wrong name.

apps/property-tree/src/features/inspections/ui/InspectionConductDialog.tsx
```diff
@@ -59,6 +82,24 @@ export function InspectionConductDialog({
...
+        includeContacts: true,
+        includeUpcomingLeases: true,
+        includeTerminatedLeases: true,
+      }),
+    enabled: !!residenceId,
+  })
+
+  const matchingLease = leases?.find(
+    (l) => l.leaseId === internalInspection?.leaseId
+  )
```
Crewiz: Could this be extracted into a dedicated hook and put into /inspection/hooks?

apps/property-tree/src/features/inspections/ui/TenantInfoCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+  tenant: TenantInfoCardData
+  address?: string
+  apartmentCode?: string | null
+}
+
+const formatDate = (value?: string | null): string => {
+  if (!value) return '-'
+  const date = new Date(value)
+  if (Number.isNaN(date.getTime())) return '-'
+  return date.toLocaleDateString('sv-SE')
```
Crewiz: You could use the existing formatISODate at shared/lib/formatters. Could add the NaN guard there aswell

apps/property-tree/src/features/inspections/ui/InspectionConductDialog.tsx
simonkropp: Ah right! Thx

apps/property-tree/src/features/inspections/ui/InspectionConductDialog.tsx
simonkropp: Yes that makes sense.. I thought about this as well but forgot to do it :D

apps/property-tree/src/features/inspections/ui/TenantInfoCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+ * Minimal view-model for the tenant card. Kept separate from the full
+ * `Tenant` type so callers can render the card from different data sources
+ * (e.g. a lightweight `TenantContactsResponse` entry or a fully-hydrated
+ * `Tenant` with its housing contract).
+ */
+export interface TenantInfoCardData {
+  contactCode: string
+  fullName: string
+  moveInDate?: string | null
+  moveOutDate?: string | null
```
Crewiz: TenantInfoCardData is defined in the component file but imported across 6 files.  Would be good to move it to features/inspections/types/index.ts since sibling types like TenantSnapshot already live there, and becauseit needs to be passed between components.

Crewiz (changes requested):

Crewiz (approved):

PR #506, opened 2026-04-13
Feature/mim 1629 hamta komponenter pa rum i besiktningsflodet

Crewiz:
Replaces the hardcoded room component list with a real GET /components/by-room/{roomId} fetch. Hardcoded surfaces (Vägg 1–4, Golv, Tak, Detaljer) act as fallbacks and are dropped when the API returns a real component with matching category (Väggar/Golv/Tak).

Test:

110-002-02-0201

check if inspection form renders tvättmaskin and torktumlare

Follow-up
[MIM-1683](https://linear.app/mimer-onecore/issue/MIM-1683) > extend InspectionRoom schema so equipment-row inspection data persists.

### DIFF
--- a/apps/property-tree/src/features/inspections/constants/components.ts
+++ b/apps/property-tree/src/features/inspections/constants/components.ts
@@ -115,6 +115,8 @@ export interface InspectionRow {
 
 // Swedish category names that supersede a default surface row when present in
 // the fetched components. Matches the existing DB convention (e.g. 'Vitvaror').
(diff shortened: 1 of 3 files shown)


simonkropp:
LGTM but I am unsure about the definition of done of this ticket? I can see that we can't save notes for the components fetched from the components endpoint. I guess this is only for equipment and not for Väggar, Golv etc? :)

simonkropp (approved):

PR #507, opened 2026-04-13
 add summary step to inspection form

Crewiz:
Stop toggle between rooms and summary, and a back link at top of summary.


apps/property-tree/src/features/inspections/ui/InspectionForm.tsx
```diff
@@ -131,58 +141,75 @@ export function InspectionForm({
...
+                </AccordionItem>
+              )
+            })}
+          </Accordion>
+        </div>
+      )}
+
+      {step === 'summary' && (
+        <div className="max-h-[70vh] overflow-y-auto pr-2 pb-24 min-w-0">
+          <button
```
simonkropp: Do we want to use a raw `<button>` component here instead of a custom component?

simonkropp (approved):
LGTM just a small comment!

PR #509, opened 2026-04-14
Feature/mim 1676 sammanstallning wire ar bostaden moblerad to form and draft

Crewiz:
Wires the "Är bostaden möblerad vid besiktningstillfället?" radio (Nej/Ja, default Nej) into the Sammanställning step and persists the value through draft-save and complete flows. Value loads back correctly when resuming a draft.

> Depends on #507 (MIM-1675 summary-step scaffolding). Please review after that one.

# Tested
- Default is "Nej" on a fresh inspection 
- Saving draft with "Nej" persists across reopen 
- Saving draft with "Ja" persists across reopen

### DIFF
--- a/apps/property-tree/src/features/inspections/hooks/useInspectionForm.ts
+++ b/apps/property-tree/src/features/inspections/hooks/useInspectionForm.ts
@@ -48,6 +48,8 @@ export function useInspectionForm(
     setInspectionTime: inspectorInfo.setInspectionTime,
     needsMasterKey: inspectorInfo.needsMasterKey,
     setNeedsMasterKey: inspectorInfo.setNeedsMasterKey,
+    isFurnished: inspectorInfo.isFurnished,
+    setIsFurnished: inspectorInfo.setIsFurnished,
 
     // Form state
     inspectionData: formState.inspectionData,
--- a/apps/property-tree/src/features/inspections/hooks/useInspectorInfo.ts
+++ b/apps/property-tree/src/features/inspections/hooks/useInspectorInfo.ts
@@ -8,12 +8,14 @@ export interface InspectorInfo {
   inspectorName: string
   inspectionTime: string
   needsMasterKey: boolean
+  isFurnished: boolean
 }
 
 export interface UseInspectorInfoReturn extends InspectorInfo {
   setInspectorName: (name: string) => void
   setInspectionTime: (time: string) => void
   setNeedsMasterKey: (value: boolean) => void
+  setIsFurnished: (value: boolean) => void
   isValid: boolean
 }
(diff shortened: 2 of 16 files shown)


apps/property-tree/src/features/inspections/hooks/useInspectorInfo.ts
```diff
@@ -39,16 +41,22 @@ export function useInspectorInfo(
     Boolean(existingInspection?.masterKeyAccess)
   )
 
+  const [isFurnished, setIsFurnished] = useState(
+    existingInspection?.isFurnished ?? false
```
simonkropp, line 45: This initializes from `existingInspection?.isFurnished` which is correct for the resume flow. However, in `InspectionFormDialog.tsx` the "start fresh" path only clears `rooms`:

```ts
userChoice === 'fresh'
  ? { ...existingInspection, rooms: null }
  : existingInspection
```

So a previously saved `isFurnished: true` would leak into a "fresh start". Should also reset `isFurnished` to `false` there, e.g.:

```ts
{ ...existingInspection, rooms: null, isFurnished: false }
```

apps/property-tree/src/features/inspections/hooks/useInspectionForm.ts
```diff
@@ -48,6 +48,8 @@ export function useInspectionForm(
     setInspectionTime: inspectorInfo.setInspectionTime,
     needsMasterKey: inspectorInfo.needsMasterKey,
     setNeedsMasterKey: inspectorInfo.setNeedsMasterKey,
+    isFurnished: inspectorInfo.isFurnished,
```
simonkropp, line 51: `MobileInspectionForm` also consumes `useInspectionForm` but doesn't destructure `isFurnished` / `setIsFurnished`, and has no summary step or furnished toggle. Currently not a compile error since the mobile save handlers are commented out, but when mobile is re-enabled it will need this wiring too. Worth a follow-up ticket or TODO so it doesn't get missed.

apps/property-tree/src/features/inspections/hooks/useInspectorInfo.ts
Crewiz: Good catch! Fixed!

apps/property-tree/src/features/inspections/hooks/useInspectionForm.ts
Crewiz: Added a todo, thinking that we should make a card to make sure the whole flow works in mobile

simonkropp (approved):
Nice! But I think you have a merge conflict ;)

PR #513, opened 2026-04-15
mim-1636: show floor plan in inspection form

simonkropp:
add button in footer to be able to open floor plan and some overall ui fixes


apps/property-tree/src/shared/ui/AlertDialog.tsx
```diff
@@ -16,7 +16,7 @@ const AlertDialogOverlay = React.forwardRef<
 >(({ className, ...props }, ref) => (
   <AlertDialogPrimitive.Overlay
     className={cn(
-      'fixed inset-0 z-50 bg-black/80  data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0',
+      'fixed inset-0 z-[90] bg-black/80  data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0',
```
Crewiz, line 19: This will affect every dialog in the app right, not just inspections? Should more primitives also have the same change from z-50 to z-[90]? Curious why it was done for some and not others 😄

apps/property-tree/src/features/inspections/ui/InspectionMoreMenu.tsx
```diff
@@ -0,0 +1,129 @@
...
+              alt="Planritning"
+              onError={() => setFloorplanError(true)}
+              className="w-full h-auto max-h-[80vh] object-contain rounded"
+            />
+          ) : (
+            <div className="flex flex-col items-center justify-center py-16 text-muted-foreground gap-3">
+              <FileImage className="h-16 w-16 opacity-30" />
+              <p className="text-sm">Ingen planritning tillgänglig</p>
+            </div>
+          )}
```
Crewiz: i think this duplicates code from ResidenceFloorplanTabsContent.tsx, could extract <FloorplanImage rentalId={...} /> in shared/ui and collapse both sites

Crewiz (approved):
Small comments, but otherwise approved 👍

apps/property-tree/src/shared/ui/AlertDialog.tsx
simonkropp: It was done because the black opaque backdrop for the dialog did not show in mobile :)

apps/property-tree/src/features/inspections/ui/InspectionMoreMenu.tsx
simonkropp: agree,.. I fixed it :)

PR #514, opened 2026-04-15
mim-1690: use only context menu when InspectionMoreMenu is used in mobile

simonkropp:
Otherwise use a regular button row..

Changes to work in https://github.com/Bostads-AB-Mimer/onecore/pull/513 :)


Crewiz:
Approved! Only thing to note is the carry-over items from #513 that still live in this file (z-index bump, duplicated floorplan render. Not really anything blocking this PR as i approved on 513 aswell.

Crewiz (approved):

PR #517, opened 2026-04-16
Feature/mim 1678 sammanstallning cost field per remark

Crewiz:
## Summary
- Adds `componentCosts` to `InspectionRoom` schema (services/inspection + core) so repair-cost estimates persist through draft save/resume
- Wires the existing "Kostnad (kr)" input in the Sammanställning table to form state via `handleComponentCostUpdate` on `useInspectionForm`

## Test plan
- [x] Manual: create inspection → mark component as Acceptabel/Skadad → open Sammanställning → enter cost → Spara utkast → reload → value persists
- [ ] No regressions in notes/status columns of the summary table


simonkropp:
The code LGTM!

There seem to be a UI bug in the summary section when I try to replace 0 and enter another number it won't allow me, so the numbers are appended to "0". However if I press the arrow keys it starts working. 

Screenshot 2026-04-21 at 13 18 57

Also if I open an old draft and try to save a new draft with some costs it throws an error. This might not be that important though, because we are actively developing and old local drafts don't need to be backwards compatible!:
Screenshot 2026-04-21 at 13 19 55

PR #522, opened 2026-04-20
add cost responsibility per component in inspection form

Crewiz:
## Summary

- Adds `componentCostResponsibilities` (`'tenant' | 'landlord' | null` per surface) to `InspectionRoomSchema` in both `services/inspection` and `core`, with factory + initial form data defaults.
- Renders a Kostnadsansvar radio group (Hyresgäst / Hyresvärd) in `ComponentInspectionCard`, between the condition buttons and the note textarea, only when condition is **Skadad** or **Acceptabel**.
- Wires the new handler through `useComponentInspection` → `useInspectionForm` → `RoomInspectionEditor` → `InspectionForm` + `MobileInspectionForm`. Legacy drafts without the field are defaulted in `useInspectionFormState` so old data keeps opening.

## Test plan

- [x] Open an inspection → room accordion
- [x] Click **God** on a component → no Kostnadsansvar row
- [x] Click **Acceptabel** → row appears, both radios empty
- [x] Select **Hyresgäst** → radio marked
- [x] Click **Skadad** → row still visible, selection persists
- [x] Click **God** → row hides
- [x] Click **Skadad** again → selection still set
- [x] Save draft → reload → selection + condition persist
- [x] Slutför besiktning → payload includes `componentCostResponsibilities` per room
- [x] Same flow on mobile form


apps/property-tree/src/features/inspections/hooks/useComponentInspection.ts
```diff
@@ -185,6 +192,29 @@ export function useComponentInspection(
     [setInspectionData]
   )
 
+  /**
+   * Update cost responsibility for a component in a room
+   */
+  const updateComponentCostResponsibility = useCallback(
```
simonkropp, line 232: **Bug: cost responsibility is not cleared when condition changes to "God"**

This new function works correctly on its own, but there's a data integrity issue in `updateCondition` above (line ~62). A user can:
1. Set condition to "Skadad"
2. Pick "Hyresgäst" as cost responsibility
3. Change condition back to "God"

The radio buttons disappear from the UI, but `costResponsibility` stays as `'tenant'` in state and gets persisted. This is stale data that could cause issues downstream.

Suggestion — in `updateCondition`, reset cost responsibility when the condition no longer warrants it:

```ts
// After computing updatedRoom.isHandled:
if (value === CONDITION_TYPE.GOOD) {
  updatedRoom.componentCostResponsibilities = {
    ...updatedRoom.componentCostResponsibilities,
    [field]: null,
  }
}
```

(You'll need to import `CONDITION_TYPE` into this hook.)

apps/property-tree/src/features/inspections/hooks/useInspectionFormState.ts
```diff
@@ -27,12 +27,23 @@ export function useInspectionFormState(
...
       // Convert array of rooms to Record keyed by roomId
-      // Ensure detailComponents defaults for rooms saved before the field existed
+      // Ensure detailComponents / componentCostResponsibilities defaults for rooms
+      // saved before those fields existed
       return existingInspection.rooms.reduce(
         (acc, room) => {
           acc[room.roomId] = {
             ...room,
             detailComponents: room.detailComponents ?? [],
+            componentCostResponsibilities:
```
simonkropp, line 56: **Nit: this default object is duplicated in three places**

The `{ wall1: null, wall2: null, ... }` literal also appears in `initialFormData.ts` and the test factory. Consider exporting a shared constant (e.g. from `initialFormData.ts` or the constants directory) so there's a single source of truth. If a new surface key is ever added you'd only update one place.

apps/property-tree/src/features/inspections/ui/ComponentInspectionCard.tsx
```diff
@@ -90,6 +101,36 @@ export function ComponentInspectionCard({
...
+      {(condition === CONDITION_TYPE.DAMAGED ||
+        condition === CONDITION_TYPE.ACCEPTABLE) && (
+        <div className="mb-3">
+          <p className="text-sm text-muted-foreground mb-2">Kostnadsansvar</p>
+          <div className="flex gap-4">
+            {(
+              [
+                COST_RESPONSIBILITY.TENANT,
+                COST_RESPONSIBILITY.LANDLORD,
+              ] as const
```
simonkropp: **Minor nit:** The `as const` on this array literal is unnecessary since both values already come from a `as const` object (`COST_RESPONSIBILITY`). TypeScript already narrows `value` to the union type in the `.map` callback without it. Not blocking, just a style thing.

simonkropp (commented):
## Review Summary

Overall a clean, well-structured change that follows existing patterns nicely. One potential data integrity issue to address, and a couple of suggestions below.

simonkropp (approved):
Tipp topp!

Just a little type check failing ;)

PR #523, opened 2026-04-20
mim-1703: Fix url linking to odoo

simonkropp:
Needed to restructure how we call odoo because of odoo 19 upgrade. Context needs to be passed differently.

Requires this change in odoo-onecore to be merged first:
https://github.com/Bostads-AB-Mimer/onecore-odoo/pull/241

How to test:

1. Go to the work order tab for building, residence, staircase etc. and press "skapa ärende" button. 
2. Check if you are redirected to odoo with the correct information pre-filled in the form.. Utrymme should be set as well as the search value for example:
Screenshot 2026-04-20 at 15 09 19


lindblomdavid:
Onecorepart looks good. but please see comment on odoo PR.

PR #527, opened 2026-04-22
Feature/mim 1683 spara besiktningsdata per komponent i besiktningsflode

Crewiz:
## Summary

Persists per-component inspection data (condition, actions, note, photos, cost, cost responsibility) for real components fetched from property-base — previously dimmed "Sparas inte än" rows. Data rides the existing `inspection.draftRooms` JSON; no DB migration.

Closes **MIM-1683**. Also completes **MIM-1674**  - sibling PR #522 shipped cost responsibility per-surface only; this PR extends it to fetched components (equipment rows + Sammanställning), which was the card's original intent.

### Changes

- **Schema.** New `InspectionComponentSchema` under `InspectionRoomSchema.components[]` with `componentId`, `label`, `condition`, `action[]`, `note`, `photos[]`, `cost?`, `costResponsibility`. Mirrored in `services/inspection` and `core`. `label` is snapshotted at write-time so summaries stay readable if property-base renames or deletes the component.
- **Hook layer.** 7 new handlers in `useComponentInspection` keyed by `componentId` — condition, action, note, add/remove photo, cost, cost responsibility. `updateComponentCondition` clears `costResponsibility: null` when the value flips back to God (mirrors the surface-keyed rule).
- **Editor.** `RoomInspectionEditor` wires equipment rows to the real handlers (including cost responsibility) and derives `isHandled` via a shared `deriveRoomIsHandled` helper — superseded surfaces drop out of the check automatically.
- **Summary (Sammanställning).** Per-component remarks now appear alongside surface remarks with the snapshotted `label`. Added 4th column "Kostnadsansvar" with a compact `<Select>` (Hyresgäst / Hyresvärd / —) for both surface and component rows. Per-component cost input is now controlled.
- **Hydration.** Defaults `costResponsibility: null` for components saved before the field existed.

### Out of scope

- Large-photo storage — inspection photos are still base64 in `draftRooms` JSON. Migration tracked in [MIM-1709](https://linear.app/mimer-onecore/issue/MIM-1709/migrate-inspection-photos-from-base64-in-draftrooms-to-file-storage).
- Writing condition/lastInspectionDate back to property-base `Components`.

## Test plan

- [ ] **Edit & save (desktop + mobile)** — on a fetched equipment row: set condition, toggle an action, enter note, capture a photo. Save draft, reopen — all data persists.
- [ ] **Cost & responsibility in editor** — set an equipment row to Acceptabel/Skadad → tenant/landlord radio appears. Pick one. Flip condition to God → radio clears and persists as `null`.
- [ ] **isHandled** — room flips to `Klar` only when every visible surface + every fetched component has a condition.
- [ ] **Summary** — Acceptabel/Skadad components appear in Sammanställning with their display name (not UUID), mixed with surface remarks. Both rows types edit cost + Kostnadsansvar inline; values round-trip through save/reload.
- [ ] **Label snapshot** — rename or delete the component in property-base after saving; summary still shows the original name.
- [ ] **Superseded surfaces** — rooms whose fetched components include a `Väggar` category hide the wall1–4 defaults; `isHandled` counts the real components.
- [ ] **Backward compatibility** — a draft saved before this PR (no `costResponsibility` on components) loads without schema errors; the field reads as `null`.
- [ ] **Finalize** — `Genomförd` transition still works.
- [ ] **CI** — `pnpm run typecheck` (18/18 green); `pnpm -F @onecore/inspection test` (105/105).

### DIFF
--- a/apps/keys-portal/src/services/api/core/generated/api-types.ts
+++ b/apps/keys-portal/src/services/api/core/generated/api-types.ts
@@ -3,9 +3,8 @@
  * Do not make direct changes to the file.
  */
 
-
 export interface paths {
-  "/auth/generatehash": {
+  '/auth/generatehash': {
     /**
      * Generates a salt and hashes the given password using that salt.
      * @description Generates a salt and hashes the given password using that salt. Pass cleartext password as query parameter.
@@ -14,18 +13,18 @@ export interface paths {
       parameters: {
         query: {
           /** @description The cleartext password that should be hashed */
-          password: string;
-        };
-      };
+          password: string
+        }
+      }
       responses: {
         /** @description Hashed password and salt */
         200: {
-          content: never;
-        };
-      };
-    };
-  };
-  "/auth/generatetoken": {
+          content: never
+        }
+      }
+    }
+  }
+  '/auth/generatetoken': {
     /**
      * Generates a JWT token
      * @description Validates username and password from request body and returns a JWT token.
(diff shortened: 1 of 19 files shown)


simonkropp:
These are not clickable for components? It seems I can't select any of them.

Screenshot 2026-04-22 at 14 34 27

simonkropp:
One more thing.. when updating the cost of the component it seems to get reset when saving draft and reloading. It works for the other costs. This should be persisted right?

Also there is still a bit of weirdness when pressing the input in order to change the value. If the value is 0 and you focus on the input the cursor is ending up after 0 so if you enter a number it becomes 0123 for example. Maybe we should just clear the input if the value is 0 and the user presses the input? Not sure if it is related to this ticket or not but it would be nice nevertheless. 

Screenshot 2026-04-22 at 14 40 53

simonkropp (changes requested):
Found some stuff while testing. Have a look if it is the same for you ;)

simonkropp (approved):
LGTM!

But we need to fix this bug on mobile ;) 
https://github.com/user-attachments/assets/171e3e42-5bf2-4794-8747-6d77e28d31e2

PR #529, opened 2026-04-24
Feature/mim 1706 hamtar fel objekt

lindblomdavid:
## Summary

Adds a `/staircases/search` endpoint to the property service, wires it into the core `/search` aggregator as a 7th result type, and surfaces staircase hits in the property-tree command palette. Also tightens the staircase schema so that the "a staircase always has a building and a property" invariant from Xpand is reflected end-to-end instead of being papered over with nullable fields in three places.

Closes MIM-1706.

## What changed

**`services/property`**
- New `searchStaircases` adapter (Prisma `findMany` on `Staircase`) with a nested `PropertyStructure` filter: `companyCode '001'`, `residenceId/localeId` null, `buildingCode/propertyCode/propertyId` not null. Placeholder codes `'00'` and `'99'` are excluded to stay consistent with the sidebar.
- New `GET /staircases/search` route behind `parseRequest({ query: staircasesSearchQueryParamsSchema })` with `q.min(3)` validation and swagger docs.
- Shared `extractPropertyData` / `mapStaircase` / `requirePropertyData` helpers deduplicate mapping between `getStaircasesByBuildingCode` and the new `searchStaircases`.
- `StaircaseSchema` split into `StaircaseBaseSchema` (id / features / dates, used by the embedded staircase on `ResidenceSchema`) and `StaircaseSchema` (base + non-null property + building).

**`core`**
- New `property-base-adapter.searchStaircases` returning `AdapterResult<Staircase[], 'staircase-search-failed'>` — specific error code rather than `'unknown'`.
- Search service aggregates staircase results alongside the existing types and registers `StaircaseSearchResultSchema` in the discriminated union. Max results bumped from 60 to 70 (10 per type × 7 types).

**`apps/property-tree`**
- Command palette renders staircase results with the `DoorOpen` icon and navigates via `paths.staircase(buildingCode, staircaseCode)`.
- Regenerated `api-types.ts` for `internal-portal`, `keys-portal`, and `property-tree`.

## Why the schema tightening is here

Before this PR, three layers (adapter, core mapper, frontend) each defended against a staircase with a null building or property. After dropping `.parse()` at the route (since the adapter now owns the invariant), there was no runtime gate catching drift. Moving the invariant into the Prisma `where` + `requirePropertyData` throw means we fail loudly at the source, the schema matches runtime, and the defensive branches in the mapper and frontend can go away.

## Test plan

- [ ] `pnpm run typecheck` — zero errors
- [ ] `pnpm -r run prettier --write` — no diff after running
- [ ] `pnpm --filter @onecore/property test staircases` — new `describe('GET /staircases/search')` passes (400 on `q < 3`, empty array on no matches, placeholder codes filtered and building present on real matches)
- [ ] Property-tree command palette: type 3+ chars matching a real staircase name → row renders with the door icon, subtitle shows building name, click navigates to the staircase page
- [ ] Hit core `/search?q=<staircase-name>` directly and verify the response includes `type: 'staircase'` entries with non-null `building.code` and `property.code`
- [ ] Verify placeholder staircases (code `00` / `99`) do **not** appear in results even when their building would match

## Out of scope

- `GET /staircases/:id` (details endpoint) — still a `//todo` in `routes/staircases.ts`.
- Migrating other search result schemas (`Property`, `Building`, `Residence`, etc.) to the specific-error-code pattern — only the new `searchStaircases` uses `'staircase-search-failed'`; the siblings still return `'unknown'`.

### DIFF
--- a/apps/property-tree/src/features/search/ui/CommandPalette.tsx
+++ b/apps/property-tree/src/features/search/ui/CommandPalette.tsx
@@ -6,6 +6,7 @@ import {
   Building2,
   CarFront,
   Command,
+  DoorOpen,
   Loader2,
   SquareUser,
   Store,
@@ -28,6 +29,7 @@ const iconMap = {
   'parking-space': CarFront,
   'maintenance-unit': Wrench,
   facility: Store,
+  staircase: DoorOpen,
 } as const
 
 function getResultProps(item: CombinedSearchResult) {
@@ -97,6 +99,19 @@ function getResultProps(item: CombinedSearchResult) {
         prefix: '[LOK]',
         subtitle: item.property?.name,
         path: paths.facility(item.rentalId),
+        state: {
+          buildingCode: item.building?.code || null,
+          propertyCode: item.property?.code || null,
+        },
+      }
+    case 'staircase':
+      if (!item.building?.code) return null
+      return {
+        icon,
+        label: item.name ?? item.code,
+        prefix: '[UPP]',
+        subtitle: item.building?.name ?? item.property?.name ?? undefined,
+        path: paths.staircase(item.building.code, item.code),
         state: {
           buildingCode: item.building?.code || null,
           propertyCode: item.property?.code || null,
--- a/apps/property-tree/src/features/search/ui/CommandPalette.tsx
+++ b/apps/property-tree/src/features/search/ui/CommandPalette.tsx
@@ -105,16 +105,15 @@ function getResultProps(item: CombinedSearchResult) {
         },
       }
     case 'staircase':
-      if (!item.building?.code) return null
       return {
         icon,
         label: item.name ?? item.code,
         prefix: '[UPP]',
-        subtitle: item.building?.name ?? item.property?.name ?? undefined,
+        subtitle: item.building.name ?? item.property.name ?? undefined,
         path: paths.staircase(item.building.code, item.code),
         state: {
-          buildingCode: item.building?.code || null,
-          propertyCode: item.property?.code || null,
+          buildingCode: item.building.code,
+          propertyCode: item.property.code,
         },
       }
     default:
--- a/services/property/src/adapters/staircase-adapter.ts
+++ b/services/property/src/adapters/staircase-adapter.ts
@@ -1,80 +1,177 @@
 import { map } from 'lodash'
+import { logger } from '@onecore/utilities'
+
 import { toBoolean, trimStrings } from '../utils/data-conversion'
 
 import { prisma } from './db'
 
 //todo: add types
 
+type StaircasePropertyData = {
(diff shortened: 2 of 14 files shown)


services/property/src/adapters/staircase-adapter.ts
```diff
@@ -1,80 +1,205 @@
...
+    ...mapStaircase(
+      staircase,
+      requirePropertyData(
+        staircase,
+        propertyStructureMap.get(staircase.propertyObjectId)
+      )
+    ),
+    buildingCode,
+  }))
+}
```
Crewiz, line 148: getStaircasesByBuildingCode is missing the try/catch + logger.error({ err }, 'staircase-adapter.getStaircasesByBuildingCode') that searchStaircases has. We need it on every adapter function so failures land in Elasticsearch with structured fields. Could you wrap it the same way?

apps/property-tree/src/features/search/ui/CommandPalette.tsx
```diff
@@ -102,6 +104,19 @@ function getResultProps(item: CombinedSearchResult) {
...
+        icon,
+        label: item.name ?? item.code,
+        prefix: '[UPP]',
+        subtitle: item.building?.name ?? item.property?.name ?? undefined,
+        path: paths.staircase(item.building.code, item.code),
+        state: {
+          buildingCode: item.building?.code || null,
+          propertyCode: item.property?.code || null,
+        },
+      }
```
Crewiz, line 118: The schema tightening guarantees building.code and property.code are non-null, but this code still uses ?. and || null fallbacks (and the early-return on 108). After narrowing on case 'staircase', you can drop the guard and use item.building.code / item.property.code directly.



```suggestion
    case 'staircase':
      return {
        icon,
        label: item.name ?? item.code,
        prefix: '[UPP]',
        subtitle: item.building.name ?? item.property.name ?? undefined,
        path: paths.staircase(item.building.code, item.code),
        state: {
          buildingCode: item.building.code,
          propertyCode: item.property.code,
        },
      }
```

Crewiz (approved):
LGTM! Some minor comments that could be fixed this PR.

apps/property-tree/src/features/search/ui/CommandPalette.tsx
lindblomdavid: agree, fix pushed . I titghetened the schemas on the other levels, after implementing it.  but didnt really alter the front end. . 

The schema tightening could be done for other endpoints as well but that is a future feature.

services/property/src/adapters/staircase-adapter.ts
lindblomdavid: fixed

PR #530, opened 2026-04-27
MIM-1712: consolidate inspection Zod schemas in @onecore/types

Crewiz:
## Summary
- Move shared inspection Zod schemas + inferred types into `libs/types/src/inspection/` following the `keys/` pattern, exposed as the `inspection` namespace from `@onecore/types`.
- Delete the duplicated `services/inspection/src/services/inspection-service/schemas.ts` and slim `core/src/services/inspection-service/schemas.ts` to only its core-only response shapes (`InspectionSchema`, `DetailedInspectionSchema`, `InspectionWithSourceSchema`, `TenantContact*`, `SendProtocol*`, `CreateInspectionRequest*`, `UpdateInspectionStatusRequest*`).
- Migrate all call sites in core and services/inspection to `inspection.X` from `@onecore/types`.


core/src/services/inspection-service/index.ts
```diff
@@ -34,8 +35,8 @@ import {
  */
 export const routes = (router: KoaRouter) => {
   registerSchema('Inspection', schemas.InspectionSchema)
-  registerSchema('InspectionComponent', schemas.InspectionComponentSchema)
-  registerSchema('InspectionRoom', schemas.InspectionRoomSchema)
+  registerSchema('InspectionComponent', inspection.InspectionComponentSchema)
+  registerSchema('InspectionRoom', inspection.InspectionRoomSchema)
   registerSchema('DetailedInspection', schemas.DetailedInspectionSchema)
   registerSchema('DetailedInspectionRoom', schemas.DetailedInspectionSchema)
```
simonkropp: Maybe not related to this PR but would be nice to fix it while we are at it?

These schemas are all registered with DetailedInspectionSchema, seems wrong?

simonkropp: DetailedInspectionRoom and DetailedInspectionRemark

services/inspection/src/services/inspection-service/index.ts
```diff
@@ -34,16 +24,28 @@ import { db } from './adapters/db'
...
-  registerSchema('XpandInspection', XpandInspectionSchema)
-  registerSchema('DetailedXpandInspection', DetailedXpandInspectionSchema)
-  registerSchema('DetailedXpandInspectionRoom', DetailedXpandInspectionSchema)
-  registerSchema('DetailedXpandInspectionRemark', DetailedXpandInspectionSchema)
+  registerSchema('XpandInspection', inspection.XpandInspectionSchema)
+  registerSchema(
+    'DetailedXpandInspection',
+    inspection.DetailedXpandInspectionSchema
+  )
+  registerSchema(
```
simonkropp, line 32: Here also 👽

simonkropp (approved):
LGTM! I just noticed some weirdness regarding registerSchema while reviewing. Maybe we should fix it while we are at it?

core/src/services/inspection-service/index.ts
simonkropp: the code highlight didn't work that well it seems

PR #531, opened 2026-04-27
sanitize ströfaktura price input to digits and decimals (MIM-1727)

Crewiz:
## Summary
Users could type spaces in the ströfaktura price field (e.g. `17 156`). `parseFloat` truncated at the first space, so Xledger received `17` instead of `17156`. The same flaw applied to commas, periods, and letters (`17,50` → `17`, `17abc` → `17`).

## Change
Sanitize the price input on every keystroke in `ArticleSection.tsx` — only digits `0-9` are accepted. Spaces, commas, periods, letters and symbols are blocked at the keystroke and never appear in the field. Whole kronor only (no decimals).

## Test plan
- [ ] Run `pnpm run dev` for property-tree, open the ströfaktura form
- [ ] In *Pris (ink. moms)*, try pressing space, `,`, `.`, letters, symbols → nothing appears in the field
- [ ] Type `17156` → field shows `17156`
- [ ] Paste `17 156` into the field → field shows `17156`
- [ ] Submit a row with price `17156`, amount `1` — verify the request body to `/invoices/miscellaneous` has `price: 17156`
- [ ] For an article with `vatExcluded: true`, confirm the VAT-excluded amount is computed from the full price

### DIFF
--- a/apps/property-tree/src/features/economy/components/ArticleSection.tsx
+++ b/apps/property-tree/src/features/economy/components/ArticleSection.tsx
@@ -27,9 +27,7 @@ interface ArticleSectionProps {
 }
 
 function sanitizePriceInput(value: string): string {
-  const stripped = value.replace(/[^\d.,]/g, '').replace(/,/g, '.')
-  const [integer, ...fractions] = stripped.split('.')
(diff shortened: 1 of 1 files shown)


henrikhenrikpersson:
Vill vi inte tillåta kommatecken eller punkter?

PR #534, opened 2026-04-28
fix bug. getLeasesForContactCode doesnt return areacode, so lease qua…

albinrm:
…lified for intern bilplats. now uses getTenantByContactCode with guards for errors


Crewiz (approved):
Approved!

PR #537, opened 2026-04-29
Feature/mim 1516 skapa besiktningsprotokoll for interna besiktningar

Crewiz:
## Summary

Mirrors the Xpand protocol flow onto internal inspections — same Protokoll dropdown
(download with/without costs, send to tenant), same PDF, same email. Closes MIM-1516.

## What changed

- **Backend**: room normalizer flattens InspectionRoom (legacy fixed keys + components[])
  into the flat remarks[] shape the renderer expects. Filter follows the canonical UI rule
  (`Acceptabel`/`Skadad` only). Four new `/inspections/internal/` routes
  (details/pdf/tenant-contacts/send-protocol). Tenant-contacts and send-protocol bodies
  extracted into shared helpers used by both sources.
- **Microservice**: `getInspectionById` was projecting only 12 of 19 inspection columns,
  so `isTenantPresent`/`notes`/etc. were silently dropped on read - the protocol PDF
  printed "Hyresgäst närvarande: Nej" on every internal inspection. SELECT and
  projection now cover all fields.
- **Shared types**: `InternalInspectionSchema` extended to match.
- **Frontend**: `inspectionService` + protocol hooks made source-aware.
  `InspectionProtocolDropdown` reused unchanged.
- **UX bonus**: "Möblerad" asked once (in summary) instead of twice; defaults to true.
  Inspection-type picker drops Inflytt, adds Underhåll (reserved for upcoming card),
  defaults to Avflytt.

## Test plan

- [ ] Create internal inspection with "Hyresgäst närvarande" ticked.
- [ ] Add reportable + non-reportable components across filter branches.
- [ ] Download both PDFs - confirm only reportable rows appear, "Möblerad: Ja",
      "Hyresgäst närvarande: Ja".


core/src/services/inspection-service/index.ts
```diff
@@ -862,6 +861,380 @@ export const routes = (router: KoaRouter) => {
...
+              pdfError,
+              errorMessage:
+                pdfError instanceof Error ? pdfError.message : String(pdfError),
+              errorStack:
+                pdfError instanceof Error ? pdfError.stack : undefined,
+              inspectionId,
+            },
+            'Error generating PDF protocol for internal inspection'
+          )
+          throw pdfError
```
simonkropp: **Suggestion (non-blocking):** Pino has a built-in serializer for the key `err` that pulls `message`, `stack`, and `type` off an `Error` automatically. Using `pdfError` as the key bypasses that serializer, which is why lines 1015–1018 have to extract `.message` and `.stack` by hand. With `{ err: pdfError, inspectionId }` you get the same fields for free and it matches the `{ err }` convention used elsewhere in this PR.

There's also a small double-logging side effect: the `throw pdfError` on line 1023 hits the outer `catch (error)` below, which logs the same failure again under a different key (`error` vs `pdfError`). Either drop the rethrow (and `return` after setting `ctx.status = 500`) or skip the inner log.

Not a blocker — the existing xpand route at line 819 has the same shape, so this is really a cleanup that could apply to both.

```suggestion
        } catch (pdfError) {
          logger.error(
            { err: pdfError, inspectionId },
            'Error generating PDF protocol for internal inspection'
          )
          throw pdfError
        }
```

simonkropp (approved):
LGTM..

One thing I was thinking about though is when we generate the PDF with costs if it should show all costs or just costs for the tenant? It seems like right now both costs for tenant and landlord gets summarised in total cost in the pdf. Worth re-visiting.

PR #542, opened 2026-05-05
Feature/avtal 106 tenfast ska kunna sla upp per kund vilken aviseringsmetod

henrikhenrikpersson:
Lägger till endpoint i economy/core för att hämta aviseringsmetod baserat på contactCode.


core/src/adapters/economy-adapter/index.ts
```diff
@@ -314,6 +314,23 @@ export async function processIMD(
...
+    {
+      contactCodes,
+    }
+  )
+
+  if (response.status === 200) {
+    return { ok: true, data: response.data.content }
+  }
+
+  return { ok: true, data: response.data.content }
```
momentiris: Im guessing we shouldn't always return ok:true?

core/src/adapters/economy-adapter/index.ts
```diff
@@ -314,6 +314,23 @@ export async function processIMD(
   }
 }
 
+export async function getInvoiceChannels(
+  contactCodes: string[]
+): Promise<AdapterResult<economy.ChannelLookup, 'unknown'>> {
+  const response = await axios.post(
```
momentiris: Why is this a POST?

core/src/services/economy-service/index.ts
```diff
@@ -293,4 +293,25 @@ export const routes = (router: KoaRouter) => {
       ctx.body = makeSuccessResponseBody(result.data, metadata)
     }
   )
+
+  router.post(
+    '/invoice-channels',
+    parseRequestBody(economy.ChannelLookupRequestBodySchema),
```
momentiris, line 367: +1 why POST?

services/economy/src/services/invoice-service/adapters/stralfors/stralfors-adapter.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (!result) {
+      if (retryCount >= config.stralfors.maxRetries) {
+        throw new Error(
+          `Failed to get a response from Strålfors API after ${config.stralfors.maxRetries} retries`
+        )
+      }
+
+      await new Promise((resolve) =>
+        setTimeout(resolve, config.stralfors.retryBackoffMs)
+      )
```
momentiris, line 129: Hmm. How blocking is this? 🤔

services/economy/test/services/invoice-service/adapters/stralfors-adapter.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    stralfors: {
+      baseUrl: 'https://stralfors.test',
+      clientId: 'test-client-id',
+      clientSecret: 'test-client-secret',
+      retryBackoffMs: 0,
+      maxRetries: 10,
+    },
+  },
+  __esModule: true,
+}))
```
momentiris: I think we already have a common mock in some base jest config file.

momentiris (commented):
Maybe i dont rly understand the post/get mix here

core/src/adapters/economy-adapter/index.ts
henrikhenrikpersson: The Strålfors API has one POST endpoint that takes up to 1000 recipients and creates an async lookup job, and one GET endpoint that returns the result of that lookup job after some time.
We don't know yet exactly how long this can take since we haven't been able to test with that many recipients, but decided to hide this polling in Onecore so that Tenfast won't need to worry about it.
We will see how it works when we have test data in Strålfors, but the idea is that it is probably faster to batch process all these recipients instead of doing it one by one (or however many contact codes we can fit in the URL if we would use a GET).

services/economy/src/services/invoice-service/adapters/stralfors/stralfors-adapter.ts
henrikhenrikpersson: I haven't been able to test with that many recipients, but it seems to take about 1 second to process ~10 in the test environment. We will see how it works when we have real test data and a bigger batch size. If it takes too long we can instead create a GET that calls the Strålfors GET and then Tenfast can handle polling on their side.

core/src/adapters/tests/economy-adapter/index.test.ts
```diff
@@ -34,4 +34,84 @@ describe('economy-adapter', () => {
...
+        error: null,
+      },
+      {
+        channel: 'eInvoiceB2C',
+        matchedCandidates: ['198112172385'],
+        error: null,
+      },
+    ]
+
+    it('returns invoice channels for given contact codes', async () => {
```
henrikhenrikpersson:
```suggestion
    it('returns invoice channels for given national registration numbers', async () => {
```

PR #546, opened 2026-05-06
Feat:AVTAL-103 Short Time Lease Adjustments

mittistormen:
* endDate is set to Lease if it is set in tenfast
* the calculation for vacantFrom skips leases with enddate that has not been terminated (terminated = cancelled is true)


momentiris (approved):
lgtm!

PR #548, opened 2026-05-08
removes filtering of non rentalobject keys on the retun loans, fetche…

albinrm:
…s their keysysteminfo, adds a tooltip that they are a not connected to the rentalobject


Crewiz (approved):
Nice

PR #551, opened 2026-05-11
Add "syncInjectedDepsAfterScripts: build" to pnpm-workspace.yaml for local build stability

svjson:
## WHAT

Added `syncInjectedDepsAfterScripts: build` to `pnpm-workspace.yaml` to make both fresh, local and CI builds behave.

## WHY

Local and fresh builds get confused and raises hard-to-diagnose errors, because:

- This workspace uses `injectWorkspacePackages: true`, which means workspace dependencies are materialized as injected package copies in pnpm’s virtual store instead of being consumed directly from the source package directory.

- That is useful/required for the Docker build/deploy setup, but it has an important consequence: when a workspace package is rebuilt, dependent packages may still compile against the previously injected copy unless pnpm syncs the injected dependency after the build.

## NOTES

-  This does not remove injectWorkspacePackages or change CI behavior. It keeps the current Docker/deploy-compatible dependency model and makes it behave correctly during both fresh and recursive workspace builds.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -8,6 +8,8 @@ packages:
 injectWorkspacePackages: true
 
 minimumReleaseAge: 5760
+syncInjectedDepsAfterScripts:
+  - build
 
 onlyBuiltDependencies:
   - '@prisma/client'


momentiris:
Why are there no actions run for this PR? Not 100% sure it applies but would be a nice reassurance seeing green builds on this

svjson:
> Why are there no actions run for this PR? Not 100% sure it applies but would be a nice reassurance seeing green builds on this

Agreed.

I think that it's security settings on the repository. I am no longer part of the Mimer team, and this PR is coming from "the outside", so CI runs are probably disabled.

It should be possible to manually start the CI runs.

momentiris:
Cant find how to do that 🙈

Crewiz:
> Cant find how to do that 🙈

Would it run if we made an empty commit?

albinrm:
im testing it here https://github.com/Bostads-AB-Mimer/onecore/tree/fixup/build-stability
 just made a clone thats not on a fork

svjson:
> Would it run if we made an empty commit?

I don't think so. The blocker is that it's a PR from a fork not under the Bostads-AB-Mimer fork.
What's missing to be able to manually trigger the runs is a workflow_dispatch trigger in ci.yaml
That needs to be present in the default branch (main), though.

albinrm:
@svjson @momentiris  it builds https://github.com/Bostads-AB-Mimer/onecore/tree/fixup/build-stability

momentiris (approved):

PR #556, opened 2026-05-12
New action on maintenance and regular return, partial return,.

albinrm:
returns the current loan but
creates a enw loan wiht teh non selected keys

concats original loan receipt and the return receipt and uses it as the new loanreceipt

packacke added for pdf concatination


albinrm:
approved i tidigare pr

PR #557, opened 2026-05-12
Can add components to component library via inspection form

Crewiz:
## Summary
- Drops `wall1`–`wall4`/`floor`/`ceiling` from `InspectionRoomSchema` and the PDF normalizer's fixed-key branch
- Inspection conduct dialog shows `+ Lägg till komponent` (nested submenus) when a surface (Vägg/Golv/Tak) is missing — populated from the seeded `Ytskikt` hierarchy - Can be added manually in prod


core/src/adapters/property-base-adapter/components.ts
```diff
@@ -535,6 +535,23 @@ async function deleteComponentModel(
   }
 }
 
+async function getSurfaceModels(): Promise<
```
simonkropp, line 538: getSurfaceModels returns err: 'unknown' for both "empty data" and any thrown error. New adapter code should not introduce more of this pattern. Distinguish at least empty vs. fetch-failure?

services/property/prisma/seed.ts
```diff
@@ -0,0 +1,219 @@
+import { PrismaClient } from '@prisma/client'
```
simonkropp: Should we rename this seed script / comment what it is seeding?

services/property/prisma/seed.ts
```diff
@@ -0,0 +1,219 @@
...
+import { components } from '@onecore/types'
+
+const prisma = new PrismaClient()
+const { SURFACE_CATEGORY_NAME } = components
+//Removing this seed for prod, will just create the compponents manually!!
```
simonkropp: What do we mean with this? :)

apps/property-tree/src/features/inspections/ui/AddSurfaceComponentMenu.tsx
```diff
@@ -0,0 +1,90 @@
+'use client'
```
simonkropp: This file has some lint warnings that are auto-fixable 🤖, also applies to InspectionSummary.tsx and RoomInspectionEditor.tsx

libs/types/src/inspection/schema.ts
```diff
@@ -129,57 +129,21 @@ export const InspectionRoomSchema = z.object({
   // their display name is resolved via the Room record from roomService.
   name: z.string().optional(),
   conditions: z.object({
```
simonkropp, line 131: Removing these will probably break old inspections in our local dev db.. but not a big deal as the feature in under development! But good to keep in mind if weird errors pop up.

simonkropp (approved):
LGTM, just some small comments

services/property/prisma/seed.ts
Crewiz: The seed is for dev/test environments, and it populates the Ytskikt hierarchy locally so the picker works out of the box. In prod we'll create those records manually via the Komponentbibliotek admin UI, or use the component-import script Albin has been working on.

PR #561, opened 2026-05-18
mim-1778: create property managers endpoint

simonkropp:
Fetches enriched keycloak users with kvartersvärd role

### DIFF
--- a/core/src/services/auth-service/index.ts
+++ b/core/src/services/auth-service/index.ts
@@ -22,6 +22,9 @@ import { requireAuth } from '../../middlewares/keycloak-auth'
  *   schemas:
  *     KeycloakUser:
  *       type: object
+ *       required:
+ *         - id
+ *         - username
  *       properties:
  *         id:
  *           type: string
@@ -33,6 +36,32 @@ import { requireAuth } from '../../middlewares/keycloak-auth'
  *           type: string
  *         email:
  *           type: string
+ *         emailVerified:
+ *           type: boolean
+ *         attributes:
+ *           type: object
+ *           description: Open-ended map of custom user attributes. Keys are realm-configurable; each value is an array of strings.
+ *           additionalProperties:
+ *             type: array
+ *             items:
+ *               type: string
+ *         createdTimestamp:
+ *           type: integer
+ *           format: int64
+ *         enabled:
+ *           type: boolean
+ *         totp:
+ *           type: boolean
+ *         disableableCredentialTypes:
+ *           type: array
+ *           items:
+ *             type: string
+ *         requiredActions:
+ *           type: array
+ *           items:
+ *             type: string
+ *         notBefore:
+ *           type: integer
  */
 export const routes = (router: KoaRouter) => {
   /**
--- a/core/src/services/auth-service/keycloak-admin-adapter.ts
+++ b/core/src/services/auth-service/keycloak-admin-adapter.ts
@@ -10,6 +10,26 @@ type GetUsersByRoleError =
   | 'role_not_found'
   | 'unknown'
 
+// Shape of a user object returned by Keycloak's admin REST API
+// (e.g. GET /admin/realms/{realm}/groups/{id}/members). `attributes` is an
+// open-ended map of custom user attributes — each value is an array of strings
+// in Keycloak, and the set of keys is realm-configurable.
+export type KeycloakUser = {
+  id: string
+  username: string
+  firstName?: string
+  lastName?: string
+  email?: string
+  emailVerified?: boolean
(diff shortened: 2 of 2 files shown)


core/src/services/auth-service/index.ts
```diff
@@ -353,4 +373,64 @@ export const routes = (router: KoaRouter) => {
       ctx.body = { content: result.data }
     }
   )
+
+  /**
+   * @swagger
+   * /auth/users/property-managers:
```
albinrm: vill vi inte hellre ha en generic get users/role?

non blocking

core/src/services/auth-service/keycloak-user-enrichment.ts
```diff
@@ -0,0 +1,193 @@
...
+ * embedding in API responses (board / list views for förvaltningsområden).
+ *
+ * - Null/undefined ids and duplicates are filtered out up front.
+ * - Returns a Map keyed by id. Missing or deactivated users map to `null` instead of
+ *   throwing, so a single bad id never crashes the response.
+ * - Uses a 60 s process-wide TTL cache. Callers should still gather all ids for one
+ *   response into a single call to dedupe per request; the helper does not maintain
+ *   any per-request context of its own.
+ */
+export async function enrichKeycloakUsers(
```
albinrm: How does this diff from the inspection usage which gets all users with a specific role?

albinrm: Ie when we get all inspectors via keycloak, seems like either we should align this one to the inspection pattern or vice versa

albinrm:
I would prefer to keep the routes role agnostic and let property-tree call it with kvartersvärd from the client side
and then

etiher 

Add the enrichment data to all user gets (ie, put phone nr and emails in the inspection consumer)

or

Add an ?enrich=true that runs the enrichment step

core/src/services/auth-service/keycloak-user-enrichment.ts
simonkropp: mobilePhone and employeeId aren't on the /roles/{role}/users response, they live on each user's profile (attributes.mobilePhone, attributes.employeeId). That's why enrichKeycloakUsers does a per-id fetch with a 60s TTL cache.. very clunky but does the job.

core/src/services/auth-service/index.ts
simonkropp: min tanke var att man kanske inte vill köra enrichment varje gång man hämtar users per roll generiskt. vi skulle kunna köra den generiska endpointen och lägga till en enrichment true/false option dock?

simonkropp:
@albinrm sounds reasonable with en enrich=true flag.

Ahh.. and one more issue is calling the generic role endpoint with åäö.. etc.. let's try to use role names that are english?

albinrm:
> Ahh.. and one more issue is calling the generic role endpoint with åäö.. etc.. let's try to use role names that are english?

i didnt think about that but probably a good idea anyways to keep the codebase name conventions an role name conventions similar, so that you know whats what at a glance

albinrm (approved):

albinrm (changes requested):
i think its probalby ok as is now but putting the question to claude outputs

the per-user lookup is avoidable. Keycloak's /groups/{id}/members endpoint accepts a briefRepresentation query parameter (defaults to true). When you pass briefRepresentation=false, the response includes the full UserRepresentation — including attributes — for every member in the same call.

so we wouldnt need the cache logic etc, maybe youve already investigaated this path?

albinrm:
also would be nice with some stricter type consistency, this is a slightly preexisting issue though

core/src/services/auth-service/keycloak-admin-adapter.ts:

has

export type KeycloakUserSummary = {
  id: string
  username?: string
  firstName?: string
  lastName?: string
  email?: string
  emailVerified?: boolean
  enabled?: boolean
}

[core/src/services/auth-service/index.ts

KeycloakUser:
  type: object
  properties:
    id, username, firstName, lastName, email, emailVerified, enabled


then we have the new zod schema for enriched user.

we should probalby have a zod for base user, another for enriched user that refernces/inherits form the base user and then use these schemas to infer types everywhere. or somethign to that effect

simonkropp:
@albinrm I looked into briefRepresentation=false and indeed this solves this issue so we don't need the enrichment function at all. Note to self: read the API docs, lol!

I figure we just return the user as is from keycloak with the full data (now that briefrepresantion is not true). Whenever a user has additional attributes they will show up in the optional "attributes" property.  We don't need to have different schemas for enriched vs unenriched users. This feels like the correct solution. :)

albinrm (approved):
Sweet

PR #564, opened 2026-05-20
Feature/mim 1803 temperatur i lagenhet pa lagenhetskort

lindblomdavid:
## MIM-1803 — Indoor temperature on the apartment card

Surfaces the latest indoor temperature for an apartment in property-tree's
"Grundläggande information", sourced from the EcoGuard Curves platform.

### What's included
- **property service**: new EcoGuard Curves integration — a dedicated
  `curves-adapter` (token caching + 401 retry), a `curves-service` that
  resolves the apartment node and aggregates per-sensor readings into
  avg/min/max buckets, and a documented `GET /apartments/{objectNumber}/temperatures`
  endpoint with unit tests.
- **core**: proxies the new endpoint through property-base (pure HTTP wrapper,
  no business logic).
- **property-tree**: a `useApartmentTemperature` hook + `ApartmentTemperature`
  component showing the average across sub-nodes at the most recent hourly
  bucket, with a link out to the EcoGuard Curves node. The fetch is
  independent and non-blocking — the rest of the card renders immediately.

### Reading semantics
Average across all sub-nodes at the latest hourly bucket, over a trailing
3-hour window. Falls back to "Uppgifter om temperatur saknas" when no reading
is available, and shows a distinct error state if the fetch fails. Note that for 99% of apartments they only have one subnode anyway. 

### Notes
- Requires the `CURVES__*` env vars (see `services/property/.env.template`)
  and `VITE_CURVES_URL` in property-tree. **(See bitwarden "onecore-property (base) local dev")**
- Generated API types regenerated across consuming apps.
- `pnpm run typecheck` passes clean across all workspaces.
- Only users with an account in curves/ecocuard can follow the link
- i didnt create a test for the front end calculation of the average temperature, felt overkill.


services/property/src/types/curves.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const ApartmentTemperaturesResponseSchema = z.object({
+  objectNumber: z.string(),
+  nodeId: z.number(),
+  from: z.number(),
+  to: z.number(),
+  interval: ApartmentTemperaturesIntervalSchema,
+  unit: z.string(),
+  series: z.array(ApartmentTemperatureSeriesSchema),
+})
```
Crewiz: Schemas shared between a service and core belong in libs/types/src/property/schema.ts as the single source of truth. (modeled after libs/types/src/keys/schema.ts)

ApartmentTemperaturesIntervalSchema, PointSchema, SeriesSchema, and ResponseSchema are currently re-defined in core/src/services/property-base-service/schemas.ts. Move them to libs/types and import via @onecore/types in both places. ApartmentTemperaturesQuerySchema isn't duplicated today but is the same concept, would keep it next to the response schema.

Not a hard blocker, but the two copies will drift if someone adds a field.

core/src/adapters/property-base-adapter/index.ts
```diff
@@ -886,6 +886,70 @@ export async function getBlockReasons(): Promise<
...
+      }
+    }
+
+    return { ok: false, err: 'unknown' }
+  } catch (err) {
+    logger.error({ err }, '@onecore/property-adapter.getApartmentTemperatures')
+    return { ok: false, err: 'unknown' }
+  }
+}
+
```
Crewiz: We want to use openapi-fetch with client().GET(...) instead of axios. the client() helper is already set up at the top of this file and 30 other endpoints use it.
This endpoint is a plain JSON GET with query params, which is exactly what openapi-fetch is built for:
```
const response = await client().GET('/apartments/{objectNumber}/temperatures', {
  params: { path: { objectNumber }, query },
})
```
The response type also comes from the generated paths/components, so the hand-written ApartmentTemperaturesResponse block above this function can be deleted entirely if that is the case.

Crewiz (approved):
Not showstoppers, but left some comments on stuff that could be  fixed for consistency 👍, otherwise approved!

services/property/src/types/curves.ts
lindblomdavid: thank you.  i will tighten my claude.md not to define twice. Seen this appear before.

core/src/adapters/property-base-adapter/index.ts
lindblomdavid: axios is still used by three other functions in the same file. (searchRentalblocks, , exportRentalBlocksToExcel, and getBlockReasons.  i agree we should use open-api fetch though. )

albinrm:
I would maybe break out the core adapters out of the main property base adapter file, (needs to be done for more adapters/routes so nothing blocking) but other than that looks good

lindblomdavid:
> I would maybe break out the core adapters out of the main property base adapter file, (needs to be done for more adapters/routes so nothing blocking) but other than that looks good

Done, see if this looks alright.

lindblomdavid:
mergar efter review och dialog med albin och sebastian

PR #567, opened 2026-05-21
Feature/mim 1795 add room during inspection writes to xpand

albinrm:


core/src/adapters/property-base-adapter/index.ts
```diff
@@ -865,6 +865,45 @@ export async function getBlockReasons(): Promise<
...
+    }
+
+    logger.error(
+      { status: response.status, body: response.data },
+      'property-base-adapter.createRoom unexpected response'
+    )
+    return { ok: false, err: 'unknown' }
+  } catch (err) {
+    logger.error({ err }, 'property-base-adapter.createRoom')
+    return { ok: false, err: 'unknown' }
```
Crewiz, line 896: Using axios here instead of the file convention of using openap-api fetch like most of the file already does. The generated types alreadycontain the POST /rooms + createRoomRequest.

I think there is 3 files that use axios.get/post but i think we want to move away from that.

services/property/src/types/room.ts
```diff
@@ -51,3 +52,44 @@ export const RoomSchema = z.object({
...
+    ) {
+      ctx.addIssue({
+        path: ['caption'],
+        code: z.ZodIssueCode.custom,
+        message: `caption "${val.caption}" is not a valid option for roomTypeCode "${val.roomTypeCode}"`,
+      })
+    }
+  })
+
+export type CreateRoomRequest = z.infer<typeof CreateRoomRequestSchema>
```
Crewiz: No need to duplicate this schema, it should live in `libs/types`. Same shape is also redefined / inlined in:

- `core/src/services/property-base-service/schemas.ts:471-510`
- `services/inspection/src/services/inspection-service/index.ts:19-21`

Crewiz (approved):
Some minor comments, but everything else looks good/works!

services/property/src/types/room.ts
albinrm: corrects should live in libs/types,
  
  but in services/inspection/src/services/inspection-service/index.ts:19-21 is a different shape

PR #568, opened 2026-05-21
Feature/mim 1684 hur ska vi lagra inspection av en komponent skick datum

Crewiz:
When an inspection is marked Genomförd, core now writes each component's condition + lastInspectionDate back to property-base via a new PUT /components/{id}/inspection-state endpoint. Per-component failures don't block completion, they're collected with the component's label and aerror message and surfaced in the inspector's UI. The schema for the write-back endpoint lives in libs/types/src/components/schema.ts so core and property-service stay in sync, and property-tree's component library now shows the latest inspection date alongside condition.

**Note on schema import style.** The new core route at `core/src/services/property-base-service/components.ts` validates the body against `components.UpdateComponentInspectionStateSchema` imported from `@onecore/types`, instead of re-declaring it in the local `schemas.ts` as every other route in the file does. This is intentional so that schemas shared between a microservice and core live in `libs/types/` (single source of truth), rather than the older convention in this file of duplicating them locally. Treat it as a small step toward that pattern, not an inconsistency.


core/src/services/property-base-service/components.ts
```diff
@@ -2015,6 +2016,96 @@ export const routes = (router: KoaRouter) => {
...
+            result.err === 'not_found'
+              ? 'Component not found'
+              : 'Internal server error',
+          ...metadata,
+        }
+        return
+      }
+
+      ctx.body = {
+        content: result.data satisfies schemas.Component,
```
simonkropp: `satisfies` is a compile-time check only — it does not strip extra fields. The adapter returns the openapi-fetch `GetComponentResponse` type, so anything property-base adds (or any drift between the two `Component` shapes) will pass straight through.

The mirror route in `services/property/src/routes/component-instances.ts` uses `ComponentSchema.parse(component)` — recommend doing the same here for runtime safety + shape narrowing.

(Same pattern exists on the other `satisfies schemas.Component[]` lines in this file — but those are out of scope for this PR; leaving them alone is fine.)

core/src/services/property-base-service/components.ts
```diff
@@ -2015,6 +2016,96 @@ export const routes = (router: KoaRouter) => {
...
+   *       required: true
+   *       content:
+   *         application/json:
+   *           schema:
+   *             type: object
+   *             properties:
+   *               condition:
+   *                 type: string
+   *                 enum: [GOOD, FAIR, DAMAGED]
+   *               lastInspectionDate:
```
simonkropp, line 2045: The Zod schema (`UpdateComponentInspectionStateSchema` in `libs/types/src/components/schema.ts`) requires both `condition` and `lastInspectionDate`, but this swagger block doesn't declare a `required: [condition, lastInspectionDate]`. Result: generated `api-types.ts` marks both as optional, which is misleading for consumers and not what the runtime actually accepts.

Add `required: [condition, lastInspectionDate]` to the schema block.

core/src/services/inspection-service/index.ts
```diff
@@ -1528,23 +1543,55 @@ export const routes = (router: KoaRouter) => {
...
+      }
+
+      // On transition to "Genomförd" (completed), write each component's
+      // condition + lastInspectionDate back to property-base. Best-effort:
+      // per-component failures are aggregated and returned alongside the
+      // inspection so the UI can surface them — the inspection itself still
+      // completes.
+      const componentWriteBackErrors =
+        body.status === 'Genomförd'
+          ? await writeBackComponentInspectionStates(result.data)
```
simonkropp, line 1574: Architectural concern: CLAUDE.md says "Core is a proxy — keep adapters as pure HTTP wrappers. No multi-step workflows or business logic." The adapters here are still pure HTTP, but the route itself now orchestrates `inspection PATCH → loop of property-base PUT`, which is a cross-service workflow.

Worth a deliberate decision: should this live inside `services/inspection` (so completing an inspection there triggers the write-back) and core stays a proxy? The current placement works, but it sets a precedent — would prefer we agree on the rule explicitly.

core/src/services/inspection-service/helpers/component-write-back.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (!inspection.rooms || inspection.endedAt === null) {
+    return errors
+  }
+
+  const lastInspectionDate = inspection.endedAt
+
+  for (const room of inspection.rooms) {
+    if (!room.components) continue
+
+    for (const component of room.components) {
```
simonkropp, line 54: Sequential `for` loop means latency is N × RTT to property-base. For inspections with many components this adds up. `Promise.allSettled` over the per-component calls would parallelize nicely and the error-aggregation pattern already fits.

Only skip this if there's a deliberate reason to rate-limit property-base.

core/src/adapters/property-base-adapter/components.ts
```diff
@@ -688,6 +688,34 @@ async function deleteComponent(
...
+
+    return { ok: false, err: 'unknown' }
+  } catch (err) {
+    logger.error(
+      { err },
+      'property-base-adapter.updateComponentInspectionState'
+    )
+    return { ok: false, err: 'unknown' }
+  }
+}
```
simonkropp, line 733: CLAUDE.md: "Return specific, semantically correct error codes — never `'unknown'`." Existing code in this file uses `'unknown'` heavily so this isn't worse, but for a new adapter it's a chance to differentiate (e.g. `bad_request` for 400, `upstream_error` for 5xx). Up to you whether to address now or leave for a sweep.

apps/property-tree/src/features/inspections/ui/InspectionConductDialog.tsx
```diff
@@ -39,6 +47,14 @@ export function InspectionConductDialog({
 }: InspectionConductDialogProps) {
   const { toast } = useToast()
   const queryClient = useQueryClient()
+  const [componentWriteBackErrors, setComponentWriteBackErrors] =
```
simonkropp: The `{ componentId; componentLabel; message }` shape is declared inline three times in this file (here, in `writeBackErrors`, and in `inspectionService.updateInspectionStatus`'s return). Easy to drift.

Two options:
1. Extract a local `type ComponentWriteBackError` and reuse.
2. Better: name the inline shape in the swagger response (`componentWriteBackErrors: { items: { $ref: '#/components/schemas/ComponentWriteBackError' } }` plus a schema definition) so openapi-typescript emits a reusable generated type.

(2) also gives the keys-portal app the same type for free if it ever needs it.

libs/types/src/property/schema.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod'
```
simonkropp, line 1: Minor convention nit: CLAUDE.md says shared schemas live at `libs/types/src/{service}/schema.ts`. `components` isn't a service name — this is really a property-service schema. Two options: move under `libs/types/src/property/`, or document why `components/` is the chosen namespace here (existing `components/surfaces` is also non-service-shaped, so there's precedent — just want it to be deliberate).

simonkropp (approved):
Nice! 

One architectural thing that I was thinking about is that we in this case need to call both property-base and inspection service from the core inspection adapter.. however I am sure we do this kind of thing in some cases. In claude.md we've stated not to do this but I don't think it matches reality.. I think that is the goal however in this case it is OK. The calls are still being handled in core API and not service to service.

Found no major issues but included some comments from claude you can look at if you like. Mostly details.

core/src/services/inspection-service/index.ts
Crewiz: I think the rule could use some clarification, so i've updated mine to:

Adapters in core/src/adapters/ are pure HTTP wrappers to one microservice — no business logic, no calling other adapters. Multi-service orchestration belongs in routes under core/src/services/.

The adapter here is a thin wrapper, and the logic lives in the route, which is what the clarified rule now says.

core/src/services/property-base-service/components.ts
Crewiz: Good catch! Did a sweep of the whole file to use .parse() instead.

core/src/services/property-base-service/components.ts
Crewiz: Done, added required: [condition, lastInspectionDate] here and in the property-service mirror route.

core/src/services/inspection-service/helpers/component-write-back.ts
Crewiz: Nice catch, switched to Promise.allSettled. Latency now ~1 RTT regardless of component count.

core/src/adapters/property-base-adapter/components.ts
Crewiz: I swept the whole file now to get it done, thanks! 👍

apps/property-tree/src/features/inspections/ui/InspectionConductDialog.tsx
Crewiz: Done 👍

libs/types/src/property/schema.ts
Crewiz: Fixed this aswell, thanks!

PR #569, opened 2026-05-22
mim-1779: show property areas/förvaltningsområden as read only board

simonkropp:
Adapted design from lovable and endpoints for /cost-centers and /cost-centers/{id}/tree.
Wired up board to data from mentioned endpoints


apps/property-tree/src/features/property-areas/data/kvvMapping.ts
```diff
@@ -0,0 +1,60 @@
+// Mappning från stewardRefNr till KVV-områdesnummer
+// Dessa nummer är 5-siffriga och representerar kvartersvärdsområden inom ett kostnadsställe
+
+export const STEWARD_TO_KVV_AREA: Record<string, string> = {
```
simonkropp: note to self: clean this mock data :D

apps/property-tree/src/features/property-areas/types/propertyArea.ts
Crewiz, line 1: PropertyAreaEntry, Steward, CostCenter, DistrictManagers have no callers, could delete the file. Guessing it is some kind of lovable leftover?

apps/property-tree/src/features/property-areas/ui/index.ts
Crewiz, line 1: I think this is also leftovers from tne initial lovable import, has no caller. Can also be deleted i think?

services/property/src/adapters/cost-center-adapter.ts
```diff
@@ -0,0 +1,192 @@
...
+          const agg = aggByCode.get(link.propertyCode)
+          return {
+            code: link.propertyCode,
+            designation: prop?.designation ?? null,
+            tract: prop?.tract ?? null,
+            addresses: agg
+              ? Array.from(agg.addresses.values()).map((b) => ({
+                  buildingCode: b.buildingCode,
+                  buildingName: b.buildingName,
+                  address: b.buildingName,
```
Crewiz: `address: b.buildingName`, these two fields will never differ. The schema implies they're distinct (building label vs. street address), so callers will pick one and break the day `bygcaption` is something like `"Hus 7"` instead of `"Mimergatan 12"`.

Either wire a real address from `cmadr`, or drop the `address` field.

services/property/src/routes/cost-centers.ts
```diff
@@ -0,0 +1,122 @@
...
+    const metadata = generateRouteMetadata(ctx)
+    try {
+      const rows = await listCostCenters()
+      ctx.body = {
+        content: rows.map((r) => CostCenterSummarySchema.parse(r)),
+        ...metadata,
+      }
+    } catch (err) {
+      ctx.status = 500
+      const errorMessage = err instanceof Error ? err.message : 'unknown error'
```
Crewiz, line 57: `reason: errorMessage` echoes raw `err.message` back to the client — leaks internals (stack traces, DB errors, file paths) on top of being the `'unknown'` pattern CLAUDE.md calls out:

> *Return specific, semantically correct error codes — never `'unknown'`.*

Return a fixed `'internal-error'` and rely on the structured log for the cause. Same fix needed at [L117-119](services/property/src/routes/cost-centers.ts#L117-L119).

apps/property-tree/src/features/property-areas/types/propertyArea.ts
simonkropp: Yes it is indeed lovable leftovers.. let me clean that up

services/property/src/routes/cost-centers.ts
simonkropp: I think we are using this return structure for all other endpoints in property-base so keeping it as is for now. I guess it might leak internals only to core API because the property-base endpoint is never called directly from the client.. so I think this is not a huge issue.

Crewiz (approved):
LGTM!

PR #571, opened 2026-05-22
No more one request per loan. Batch and orchestration!

albinrm:


apps/keys-portal/src/services/api/contactService.ts
```diff
@@ -23,6 +23,73 @@ export async function fetchContactByContactCode(
...
 }
 
+/**
+ * Batch lookup of contacts by contact code through the v1 contacts API.
+ * Lean by default — pass include flags to opt in to phone/email/address joins.
+ *
+ * Missing contact codes are simply absent from the response — the caller
+ * should treat a missing entry as "not found" rather than an error.
+ */
+export async function fetchContactsByContactCodeBatch(
```
Crewiz: This is unused. Is the plan to use it or can it be removed?

Crewiz (approved):
LGTM!

core/src/api/v1/contacts/index.ts
```diff
@@ -127,6 +128,63 @@ export const routes = (router: OkapiRouter, config: Config) => {
...
+        includePhone,
+        includeEmail,
+        includeAddress,
+      })
+
+      if (response.ok) {
+        ctx.status = 200
+        ctx.body = {
+          ...metadata,
+          content: { contacts: transformContacts(response.data) },
```
momentiris: There is a `makeSuccessResponseBody` export from utilities

albinrm: fixed by aligning with the type efined in the other list endpoints and using their helper taht incorporates makeSuccessResponseBody. good catch

apps/keys-portal/src/services/api/contactService.ts
albinrm: I used it previously for single call lookups to the new batch endpoint. Then I migratied away formt this to the core orchestration funciton so it lost all its callers. WIll remove as it can easily be added back if needed. 

The core standalone batch endpoint now also doesnt have any real callers but I will leave that, seems useful to have ready to ogp.

PR #573, opened 2026-05-25
removes the green box after feedback it was just confuding

albinrm:
feedback fr[n ekonomi och kc att denna ska bort
image


Crewiz (approved):
🥇

PR #574, opened 2026-05-25
removes filter of AR type of invoiceEventrs (basically the invoiceeve…

albinrm:
…nts that represent the creation of the actual invoices. if type is credit


albinrm:
partner till https://github.com/Bostads-AB-Mimer/onecore/pull/575

momentiris (approved):

PR #576, opened 2026-05-26
Epic/mim 829 epic lagenhetsbesiktning

Crewiz:
## Summary

Launches the complete apartment inspection (besiktning) flow into main.

This PR adds the inspector-facing workflow end-to-end: per-component condition data and dates, room and component creation during inspection, floor plan visualization, tenant information, contract picker, and the summary step (sammanställning) with cost responsibility per remark and component. Photos are now stored via `file-storage` instead of the legacy path, inspection Zod schemas are consolidated in `@onecore/types` so the microservice and core share one source of truth, and the mobile UX has been brought in line with the desktop flow. PDF protocols can be generated for internal inspections (with or without costs).

## Test plan

- [ ] All CI checks green (typecheck, build, prettier, tests)
- [ ] Create + complete a new internal inspection from `/besiktningar` (desktop + mobile)
- [ ] Add room and component during inspection
- [ ] Generate PDF protocol both with and without costs
- [ ] Resume a draft inspection and submit
- [ ] Keycloak users with the `besiktning` role can access the full flow


simonkropp (commented):
Maybe remove findings-mim1732.md ? Seems like an AI generated document.

PR #578, opened 2026-05-26
AVTAL-104: Contacts role

momentiris:
Adds role based access to v1 contacts read routes

Added contacts:read role (to staging keycloak) and assigned it to a new tenfast client per this wiki:
https://dev.azure.com/mimeronline/ONECore/_wiki/wikis/Digital-verksamhetsutveckling.wiki/201/Keycloak-service-clients-med-beh%C3%B6righetskontroll

This role doesn't exist in keycloak prod. AFAIK shouldn't be a problem since `requireRole` is "or"-based


mittistormen (approved):
Ser bra ut!

PR #579, opened 2026-05-27
mim-1786: adapt old kvv-district usage to new model

simonkropp:
Rewiring the app to use the new onecore_* tables containing cost-centers etc instead of fetching this from xpand. Also the property-managers should be fetched and managed in keycloak.

Had to refactor a bit so that existing filters and searches and add kvv-areas endpoint that fetches förvaltningsområde from keycloak id.


Crewiz (approved):
LGTM!

albinrm:
Run prettier on the generated files, seems like youre adding ; and "  everywhere on those, hence the 10k changes

PR #581, opened 2026-05-27
mim-1810: bugfix: seed cost center "code" correctly - was seeding code with same value as name

simonkropp:
Seed was setting cost center "code" to the same value as "name". Should have a unique numeric code like "61130" for example..


Crewiz (approved):
LGTM

PR #587, opened 2026-05-27
Feature/mim 1780 feature byt ansvarig kvartersvard for kvv omrade

Crewiz:
Implements the "change responsible" flow for KVV areas (förvaltningsområden board).

### Changes
- **`services/property`**: new `PATCH /kvv-areas/:id/responsible` — updates `responsible_keycloak_user_id` + audit columns via Prisma.
- **`core`**: new route at the same path. Gated by `property-areas:write` (MIM-1788). Validates the target user is in the `property-manager` Keycloak group before forwarding to the property service.
- **`apps/property-tree`**: pencil icon on each KVV-area board column header → modal with a Keycloak-user dropdown sourced from `usePropertyManagers`. 

### Test plan
- [x] Happy path: swap kvartersvärd on two columns, persists across reload
- [x] Permission gating: pencil disabled when caller lacks `property-areas:write`
- [x] Cancel path: no request fires

#
- `employeeId` may render blank for some users in the dropdown — that's [MIM-1812](https://linear.app/mimer-onecore/issue/MIM-1812) (Azure → Keycloak attribute mapping), **not introduced by this PR**.
- Follow-up candidate (not blocking): `services/property/src/types/kvv-area.ts` Zod schemas would ideally live in `libs/types/src/property/` per the shared-types convention. Maybe this could be a whole sweep of property, a separate ticket.

Linear: [MIM-1780](https://linear.app/mimer-onecore/issue/MIM-1780)


albinrm:
my bad g mer konflikter

simonkropp (approved):
Looks good to me and works :)

simonkropp (changes requested):
I think there might be an issue - the pen edit button seems to appear even if I don't have permissions to edit. I believe it should only show if you have property-areas:write role?

I also seem to be able to change the property-manager in the BE regardless of my role.

simonkropp:
The pen button seem to be grayed out when logging out and logging in again so maybe it works sort of.. ?

PR #588, opened 2026-05-28
mim-1812: sync entra id attributes to keycloak script

simonkropp:
Created a script that tries to pull each user from azure/entra id and then updates the keycloak user with attributes
Mapping the attributes through mappers was tried and ended up being more cumbersome then expected because some attributes are not returned by default. The easiest solution is to pull this from the graph api manually. 

Following secrets needs to be added to core:
MICROSOFT_GRAPH__TENANT_ID
MICROSOFT_GRAPH__CLIENT_ID
MICROSOFT_GRAPH__CLIENT_SECRET

Also requires onecore client in keycloak to have service account roles manage-users in order to update keycloak users


albinrm (approved):
🔥

PR #596, opened 2026-06-01
Service and core PUT route for kvv areas. frontend implementationwith…

albinrm:
… drag and drop, override state handling for the trees. NOTE no role restrictions on the calls yet

### DIFF
--- a/apps/property-tree/src/pages/PropertyAreasPage.tsx
+++ b/apps/property-tree/src/pages/PropertyAreasPage.tsx
@@ -410,14 +410,7 @@ export function PropertyAreasPage() {
         ) : isMobile ? (
           <StewardAdminMobile
             kvvAreas={kvvAreaList}
-            propertiesByKvvArea={
-              new Map(
-                kvvAreaList.map((k) => [
-                  k.kvvArea,
-                  propertiesByKvvArea.get(k.kvvAreaId) ?? [],
-                ])
-              )
-            }
+            propertiesByKvvArea={propertiesByKvvArea}
           />
         ) : (
           <div className="grid grid-cols-[minmax(0,1fr)] flex-1 min-h-0">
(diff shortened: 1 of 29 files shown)


core/src/adapters/property-base-adapter/index.ts
```diff
@@ -284,6 +284,50 @@ export async function getPropertyDetails(
...
+  propertyCode: string,
+  body: { kvvAreaId: string; updatedBy?: string | null }
+): Promise<
+  AdapterResult<
+    PropertyKvvAreaLink,
+    'property-not-found' | 'kvv-area-not-found' | 'unknown'
+  >
+> {
+  try {
+    const response = await axios.put(
```
simonkropp: I think mistakenly axios is used here instead of openapi-fetch like the other endpoints?

apps/property-tree/src/pages/PropertyAreasPage.tsx
```diff
@@ -263,21 +410,54 @@ export function PropertyAreasPage() {
         ) : isMobile ? (
           <StewardAdminMobile
             kvvAreas={kvvAreaList}
-            propertiesByKvvArea={propertiesByKvvArea}
+            propertiesByKvvArea={
+              new Map(
```
simonkropp: note from mr. claude: builds a new Map on every render to translate kvvAreaId → kvvArea for the mobile component. Minor; either useMemo or migrate StewardAdminMobile to be keyed by id too.

simonkropp (approved):
Good job! Also works really well.. I liked the edit functionality where you have to confirm the edits.. prevents accidentally moving stuff. 

One thing you should fix before merging though is the usage of axios directly - it is a departure from what endpoints typically use (openapi-fetch).

Otherwise feel free to merge :)

core/src/adapters/property-base-adapter/index.ts
albinrm: ye should be openapi-fetch

apps/property-tree/src/pages/PropertyAreasPage.tsx
albinrm: ill have a look

PR #608, opened 2026-06-05
Feature/mim 1856 byt sms fran standard infobip till tele2 infobip i

lindblomdavid:
Tele2 procurement. Tele2 runs the same Infobip SMS platform (identical
/sms/3/messages contract and `App` auth), so only the base URL and API
key change — sender "Mimer" and request shape are unchanged. Email stays
on Infobip.

- Add tele2 + health.tele2 config (TELE2__BASE_URL / TELE2__API_KEY)
- Point sendSmsV3 and its logs at config.tele2
- Add tele2SmsHealthCheck and wire it into the health service
- Add TELE2 vars to .env.test

Added env variables to production in lens and in operations.

### DIFF
--- a/apps/property-tree/src/entities/release-note/data/2026-06-05-v3.5.1.json
+++ b/apps/property-tree/src/entities/release-note/data/2026-06-05-v3.5.1.json
@@ -1,0 +1,9 @@
+[
+  {
+    "id": "rn-2026-06-05-v351-1",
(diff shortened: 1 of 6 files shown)


lindblomdavid:
ok i slack av albin. 

Har även lagt in variabler i kubernetes. (produktion)

PR #622, opened 2026-06-15
Feature/mim 1829 persist outbound communications dispatch message recipient

albinrm:


services/communication/src/services/communication-log-service/adapters/db.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  return { dispatch, recipients }
+}
+
+/**
+ * Per-customer communication timeline: one row per message_recipient owned
+ * by `kundId`, joined to its dispatch. Newest first. Direction-agnostic —
+ * once inbound logging is added, replies surface here next to outbound
+ * messages without query changes.
+ */
```
simonkropp, line 138: Got this suggestion from Claude.. (Maybe there's a reason why the sorting is in JS and not SQL.. anyways leaving this here):

Consider doing the join + sort in SQL instead of fanning out into two queries and stitching them in JS.

Right now this issues a `SELECT FROM message_recipient WHERE kundId = ?`, then a second `SELECT FROM dispatch WHERE id IN (...)`, then builds a Map, filters out orphans, and sorts by `triggeredAt` in JS. For a customer with a large history that's two round-trips and an O(n log n) sort on the app server — and the orphan filter masks data-integrity issues (the FK already prevents that case).

A single query keeps it indexed and lets MSSQL do the sort:

```ts
return db('message_recipient as r')
  .join('dispatch as d', 'd.id', 'r.dispatchId')
  .where('r.kundId', kundId)
  .orderBy('d.triggeredAt', 'desc')
  .select({
    'recipient.id': 'r.id',
    'recipient.dispatchId': 'r.dispatchId',
    // ...etc, or use two selects and remap
  })
```

Or more cleanly, select the columns flat and remap once:

```ts
const rows = await db('message_recipient as r')
  .join('dispatch as d', 'd.id', 'r.dispatchId')
  .where('r.kundId', kundId)
  .orderBy('d.triggeredAt', 'desc')
  .select('r.* as recipient', 'd.* as dispatch') // pseudo — use explicit aliases
```

The existing `idx_message_recipient_kundId_dispatchId` index covers the lookup side. Worth doing now while volumes are small; the JS sort will be the first thing to hurt once a power user has thousands of messages.

core/src/adapters/communication-adapter/log-reads.ts
```diff
@@ -0,0 +1,42 @@
+import { loggedAxios as axios } from '@onecore/utilities'
+import { communication } from '@onecore/types'
+
+import config from '../../common/config'
+import { AdapterResult } from '../types'
+
+export const getCustomerMessages = async (
+  kundId: string
```
simonkropp, line 8: Do we want to use a swedish variable kundId here.. or should we call it customerId or similar?

services/communication/src/services/infobip-service/adapters/email-adapter.ts
simonkropp, line 10: BulkEmail unused: can be removed

services/communication/src/api.ts
```diff
@@ -1,12 +1,25 @@
...
+// package.json keeps the two router type packages from conflicting.
+const legacyRouter = new KoaRouter()
+infobipRoutes(legacyRouter)
+healthRoutes(legacyRouter)
+linearRoutes(legacyRouter)
 
-infobipRoutes(router)
-healthRoutes(router)
-linearRoutes(router)
+const okapi = makeOkapiRouter(new KoaRouter(), {
```
simonkropp, line 18: Svens router? We want to start using this? :)

services/communication/.env.template
```diff
@@ -9,4 +9,10 @@ LINEAR__URL=https://api.linear.app/graphql
 LINEAR__API_KEY=
 LINEAR__TEAM_ID=19b92370-c7b2-44ec-a26c-0f067edc7070
```
simonkropp, line 10: these linear values, do we want to have them in the template file?

simonkropp (commented):
LGTM!

core/src/adapters/communication-adapter/log-reads.ts
albinrm: good point...

services/communication/.env.template
albinrm: its a little questionable yeah, but its preexisting

services/communication/src/api.ts
albinrm: yea, when making new services i have been using it. This isnt technically a new service but there wasnt any pre-existing swagger in this one.

services/communication/src/services/communication-log-service/adapters/db.ts
albinrm: Ive concidered this, i defered it because it will make the code a fair bit wordier, but yes it should be done for efficiency

services/communication/src/services/infobip-service/adapters/email-adapter.ts
albinrm: lookign into it

services/communication/.env.template
albinrm: its fine, without the api key  its still inaccessible, still a bit questionable

services/communication/src/services/communication-log-service/adapters/db.ts
albinrm: created a s eparate issue for this, i cna get ion that one when im back from midsummer :P

services/communication/src/services/infobip-service/adapters/email-adapter.ts
albinrm: done

core/src/adapters/communication-adapter/log-reads.ts
albinrm: that was a suprisingly big refactor... done

PR #623, opened 2026-06-16
AVTAL-135: adds förvaltare and huvudman to get contact endpoints

johanneskarlsson:


services/contacts/src/adapters/xpand/repository.ts
```diff
@@ -77,7 +83,17 @@ export const xpandContactsRepository = (
...
+      if (!contact) return null
+      // Only individuals participate in guardianship relations; skip the two
+      // extra queries for organisations (relatedContacts already defaults to []).
+      if (contact.type === 'individual') {
+        contact.relatedContacts = await relatedContactsFor(
+          db.get(),
+          contact.contactCode
+        )
+      }
+      return contact
```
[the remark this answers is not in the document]
johanneskarlsson: I will remove that guard. Orgs might have relations, I don't think so, but they might so I will include relatedContacts on those single GET methods too.

services/contacts/.jest/sql/seed.sql
```diff
@@ -292,3 +292,19 @@ VALUES
   ('_0J415P786     ', 'adrfakt        ', '_RQA11RNMA     ', '_0J4158OO1     ', 'SE', 'Testgatan 2', NULL, '722 22', 'VÄSTERÅS', 'SVERIGE', NULL, NULL, NULL, NULL, NULL, NULL, NULL, '1x-415P787')
   -- END cmadr ROWS
 ;
+
+-- Guardian relations (AVTAL-135).
+-- P000555 has förvaltare P000444 (forvtyp = 2).
+-- P000666 has god man P000444 (forvtyp = 1) — exists to prove the
+-- förvaltare endpoints filter on type.
+UPDATE cmctc SET keycmctc2 = '_0J4157DDD     ', forvtyp = 2 WHERE cmctckod = 'P000555';
+UPDATE cmctc SET keycmctc2 = '_0J4157DDD     ', forvtyp = 1 WHERE cmctckod = 'P000666';
```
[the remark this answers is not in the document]
johanneskarlsson: Yes, trustee endpoint currently returns both God man and förvaltare since the forvtyp is not set. I will fix that in another ticket though.

services/contacts/src/adapters/xpand/guardian-query.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const toRelatedContact = (
+  row: RelatedRow,
+  role?: RelatedContactRole
+): RelatedContact | null => {
+  const contactCode = row.contactCode?.trim()
+  if (!contactCode || contactCode === REDACTED_CONTACT_CODE) return null
+  return {
+    contactCode,
+    role: role ?? ROLE_BY_FORVTYP[row.forvtyp as number],
```
momentiris: what if forvtyp is null?

services/contacts/src/adapters/xpand/guardian-query.ts
```diff
@@ -0,0 +1,156 @@
...
+        db.raw('?', [ADMINISTRATOR_FORVTYP])
+      )
+    })
+    .whereRaw('TRIM(subject.cmctckod) = ?', [contactCode.trim()])
+    .select(...RELATED_COLUMNS)
+
+  return {
+    subjectExists: rows.length > 0,
+    related: toRelatedContacts(rows, 'administrator'),
+  }
```
momentiris, line 83: Nit pick maybe but feels like this subjectExists boolean could be replaced by `related.length` at call site

services/contacts/src/adapters/xpand/guardian-query.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (contactCodes.length === 0) return result
+
+  const trimmed = contactCodes.map((c) => c.trim())
+  const placeholders = trimmed.map(() => '?').join(', ')
+  const inSubjectCodes = `TRIM(subject.cmctckod) IN (${placeholders})`
+
+  const forwardRows = await db('cmctc as subject')
+    .innerJoin('cmctc as related', 'subject.keycmctc2', 'related.keycmctc')
+    .whereRaw(inSubjectCodes, trimmed)
+    .whereIn('subject.forvtyp', [1, 2])
```
momentiris: 1, 2 hardcoded. Looks like we have a constant for those?

services/contacts/src/adapters/xpand/guardian-query.ts
```diff
@@ -0,0 +1,156 @@
...
+  const push = (
+    subjectCode: string | null,
+    row: RelatedRow,
+    role?: RelatedContactRole
+  ) => {
+    const key = subjectCode?.trim()
+    const related = toRelatedContact(row, role)
+    if (!key || !related) return
+    if (!result.has(key)) result.set(key, [])
+    result.get(key)!.push(related)
```
momentiris: I'd void the use of "!". Red flag IMO

services/contacts/src/adapters/xpand/repository.ts
```diff
@@ -77,7 +83,13 @@ export const xpandContactsRepository = (
         .hasContactCode(contactCode)
         .getOne(db.get())
 
-      return transformDbContactRows(dbContactRows)[0]
+      const contact = transformDbContactRows(dbContactRows)[0]
+      if (!contact) return null
+      contact.relatedContacts = await relatedContactsFor(
+        db.get(),
+        contact.contactCode
+      )
```
momentiris, line 110: This means getting contact by code always queries relations? The list one below looks optional

services/contacts/test/e2e/contacts.byContactCode.test.ts
```diff
@@ -58,6 +58,7 @@ describe('/contacts/:contactCode', () => {
           fullName: 'Fiktivsson Exempel',
```
momentiris, line 58: This file needs a test on when relations are included/excluded. I think they are always included right now

services/contacts/src/adapters/xpand/repository.ts
johanneskarlsson: Yes. The batch endpoint is the only one where relations are optional. A bunch of other stuff is also optional on that endpoint which is why chose to do the same for relations 🙂

services/contacts/src/adapters/xpand/guardian-query.ts
johanneskarlsson: No, since we are left joining, `related.length === 0 ` is true for both "contact doesn't exist → 404" and "contact exists but has no förvaltare → 200 []" 🙂

momentiris (approved):

PR #629, opened 2026-06-16
fix: sort property manager filter list 

simonkropp:
Now it sorts alphabetically.. :)


albinrm (approved):
i dont think you even needed to have this reviewed :P

PR #633, opened 2026-06-22
Fix: /leasing/batch endpoint returns double leases

mittistormen:
when there is multiple invoice recipients. This is a bug! The endpoint doesn't even return any info on the recipient.

* Query now selects hyobj directly instead of joining in from hyavk


johanneskarlsson (approved):
🙌

PR #643, opened 2026-06-24
Feature/mim 1734 sortera soksvar i ordning

Crewiz:
## Sort search results by object number (MIM-1734)

Search results were returned in arbitrary order, making objects hard to find. This sorts all search endpoints by object number ascending.

### Changes
- **6 search adapters** (buildings, facilities, maintenance units, parking spaces, properties, staircases) — added DB-side `orderBy ... 'asc'`.
- **Residence search** — sorted in-memory by `rentalId`, since it comes from a nested to-many relation that Prisma can't order directly.

Object numbers are zero-padded fixed-width, so a plain lexicographic sort yields correct numeric order, consistent across all seven endpoints.

Closes MIM-1734

### DIFF
--- a/services/property/src/adapters/building-adapter.ts
+++ b/services/property/src/adapters/building-adapter.ts
@@ -189,6 +189,7 @@ const searchBuildings = async (
             },
           },
         },
+        orderBy: { buildingCode: 'asc' },
         take: 10,
       })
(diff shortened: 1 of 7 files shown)


albinrm (approved):
if the performance is still good then i dont see any issue

PR #648, opened 2026-06-25
Feature/mim 1831 infobip delivery webhooks message recipient status

Crewiz:
## MIM-1831: Infobip delivery webhooks → message_recipient status

image
image

Persisted message status never moved past `sent`. This adds `POST /webhooks/infobip`
in services/communication that consumes Infobip delivery reports and updates
`status` / `statusUpdatedAt` / `error`, lighting up the "Levererat / Misslyckades /
Studsade" badge in the tenant Kommunikationslogg tab. Closes MIM-1831.

**Email** delivery reports arrive via the Infobip account Subscription (HTTP Basic).
**SMS** runs through Tele2's separate Infobip account that we don't
control, so reports are requested per-message with the secret in the URL (`?token=`).
This token approach is interim,  once Tele2 (or we get access) sets up an account-level subscription
like email's, SMS can converge on the same Basic/subscription mechanism.

The webhook secret only authorizes flipping a message's delivery status (and the
displayed provider error) .no data access, no destructive action; worst case is a
mislabeled status badge.

Verified live end-to-end on both channels (delivered, failed, and bounced).

### Rollout
- Ops: ensure `/webhooks/infobip` is publicly reachable over HTTPS, same as the
  existing keys `/webhooks/simplesign` exposure.
- Set on the communication service (auth is skipped if all unset):
  - Email (Basic): `INFOBIP__WEBHOOK_USERNAME`, `INFOBIP__WEBHOOK_PASSWORD`
  - SMS (token):   `INFOBIP__WEBHOOK_TOKEN`, `INFOBIP__SMS_DELIVERY_REPORT_URL`
- Email only: point the Infobip Notification profile's Webhook URL at the prod URL
  (Basic setting carries over). SMS needs no portal change, the callback URL + token
  are sent per-message by code.


albinrm (changes requested):
Already in dms but listing for posterity.

1 Keycloak for basic auth
2 Auth (even the new URL based auth for tele2) in core
3 DeliveryReport not optional
4 Config CORE/BASE__URL  instead of INFOBIP__SMS_DELIVERY_REPORT_URL then use CORE/BASE__URL + /webhooks...... in the implementation



EDIT:  

1 is adressed
2. SMS should go through core as well, incl moving the token handling there. 
Not sure if communications routes are publically exposed on the ingress? they shouldnt be so i dont think it would be reachable at all on our epic or prod environments as is
3: is adressed
4: not urgent, but we will need to point it at core rather than communication

albinrm:
also point this pr to the epic :D

PR #654, opened 2026-07-01
make contactCode required through the single SMS/email send flow

albinrm:
Prevents unattributed communication-log rows: single sends could reach /sendBulkSms|/sendBulkEmail with a null contactCode, so the dispatch never attached to the customer's timeline.

- Remodel useSingleSms/useSingleEmail state as `recipient | null` (a fully required recipient) instead of a flat object seeded with empty strings, so there is no "open but blank" state that can be sent — openness is derived from having a recipient.
- Send via React Query useMutation (toast + tenant-communication invalidation in onSuccess/onError, isPending for free) instead of hand-rolled async, matching the mutation pattern used across the app.
- Require contactCode on tenantService.sendBulkSms/sendBulkEmail and the useBulkMessaging option types, so every send path (single + bulk) must supply it at compile time.
- Update the three call sites (TenantPage, TenantsTabContent, CurrentTenant) to open with a full recipient object and render the modal only when a recipient exists.

### DIFF
--- a/apps/property-tree/src/features/tenants/ui/CurrentTenant.tsx
+++ b/apps/property-tree/src/features/tenants/ui/CurrentTenant.tsx
@@ -81,37 +81,45 @@ export function CurrentTenant({
               <TenantLeaseCard
                 tenant={tenant}
                 onSendSms={(phone) =>
-                  sms.openSmsModal(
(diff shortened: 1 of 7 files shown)


Crewiz (approved):
👍

PR #656, opened 2026-07-01
log work-order tenant SMS/email to the communication log

albinrm:
Tenant-facing work-order messages are dispatched from Odoo through /sendWorkOrderSms and /sendWorkOrderEmail, but were never recorded in the communication log. Log them at those routes the same way the parking offer flow does, so they surface on the customer timeline.

Thread optional contactCode and triggeredByUser end-to-end (Odoo -> core routes -> communication-adapter -> send routes). Both are TEMPORARILY optional: Odoo deploys independently of OneCore, so older callers don't send them yet. While optional, the route only logs when contactCode is present (no contactCode -> no log row, no behavior change). A TODO marks where to make contactCode required and drop the guard once every Odoo caller sends it

### DIFF
--- a/core/src/adapters/communication-adapter/index.ts
+++ b/core/src/adapters/communication-adapter/index.ts
@@ -268,6 +268,8 @@ export const sendWorkOrderSms = async ({
   phoneNumber,
   text,
   externalContractorName,
+  contactCode,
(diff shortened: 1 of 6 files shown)


Crewiz (approved):
👍

PR #658, opened 2026-07-02
feat: search Odoo errands by number from top search (MIM-1874)

Crewiz:
## MIM-1874 — Söka på ärende från toppsök

Lets kundcenter/kvartersvärdar/bosociala jump straight to an Odoo errand
from the property-tree top search. Type `od-12345` (or `od-1`) and get a
validated **[ÄRENDE]** result that opens the errand in Odoo in a new tab.

### How it works
- **New lookup-by-id path:** work-order service `GET /workOrders/id/:id`
  → core `GET /work-orders/by-code/:code`. The `od-` code is just the Odoo
  record id, so no onecore-odoo changes were needed.
- **Detection:** an errand search must start with `od-` (`parseErrandNumber`).
  This disambiguates from other numeric input and clears the 3-char minimum.
- **Search behavior:** an `od-…` query hits **only** the errand endpoint —
  the property/contact backends aren't touched. Non-errand queries never hit
  Odoo. Each source is failure-isolated, so a slow/down source degrades to
  empty instead of blanking the whole palette.

### Testing
Run onecore + odoo locally. Search a real errand id as `od-<id>` → row appears → click
opens Odoo. A non-existent id correctly shows nothing.

### DIFF
--- a/core/src/services/work-order-service/index.ts
+++ b/core/src/services/work-order-service/index.ts
@@ -1451,7 +1451,7 @@ export const routes = (router: KoaRouter) => {
         'Error getting work order by code'
       )
       ctx.status = 500
-      ctx.body = { error: result.err, ...metadata }
+      ctx.body = { error: 'Internal server error', ...metadata }
     } catch (error) {
       logger.error(error, 'Error getting work order by code')
       ctx.status = 500
--- a/core/src/services/work-order-service/index.ts
+++ b/core/src/services/work-order-service/index.ts
@@ -1356,6 +1356,111 @@ export const routes = (router: KoaRouter) => {
 
   /**
    * @swagger
+   * /work-orders/by-code/{code}:
+   *   get:
+   *     summary: Get a single work order by its errand code
+   *     tags:
+   *       - Work Order Service
+   *     description: >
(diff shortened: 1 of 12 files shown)


core/src/services/work-order-service/index.ts
lindblomdavid, line 1: Leaking 'unknown' as the client-facing error string — [core/.../work-order-service/index.ts](vscode-webview://0j6e8pb8lt0e4e9tdggf9l9pp2p0sqeto7u0igne9sffabpnrio9/core/src/services/work-order-service/index.ts)


ctx.status = 500
ctx.body = { error: result.err, ...metadata }   // result.err === 'unknown'
This sends the literal "unknown" to the browser.

lindblomdavid:
The od- prefix is stripped in three places. The frontend parseErrandNumber returns digits only (match[1]), then the core adapter strips ^od- again (a no-op on already-bare input), then the downstream service does Number(). It works, and the defensive stripping is why your "bare numeric code" test passes — but it's redundant layering. Not worth reworking now; just be aware the contract is "any layer accepts either form," which is slightly more surface area than needed. The comments do a good job documenting each layer, so this is a nit.

PR #665, opened 2026-07-06
MIM-1893: Skapa ströfaktura fran kundkortet

henrikhenrikpersson:


albinrm (changes requested):
Main feature works perfectly as intended and could be merged. Consider making some minor changes (non-blocking):

Minor change, but I would like to move the button for this to the invoices tab. Adding actions directly on the tenant card will quickly make it cluttered

Suggestions from claude that I agree with:

1. Gate the "Laddar kund..." indicator so it doesn't flash on manual select

Since we now mirror the selected contactCode into the URL, picking a tenant from the search box re-triggers useTenant, which flips isLoadingTenant true and flashes "Laddar kund..." even though we already have the name. Gate the loading text on there being no selection yet, so it only shows during the URL-driven initial load:


// MiscellaneousInvoiceForm.tsx (~line 397)
tenantName={
  isLoadingTenant && !selectedTenant ? 'Laddar kund...' : selectedTenant?.fullName
}

Same reasoning for the not-found toast (~line 182) — add && !selectedTenant to its condition so a failed re-fetch of an already-selected tenant can't fire a false "Kund kunde inte hittas" and clear the param:

if (contactCodeFromUrl && !isLoadingTenant && !tenant && !selectedTenant) {


2. Extract a symmetric param helper

clearContactCodeParam is extracted but the set path is inlined in handleSelectTenant, duplicating the new URLSearchParams(prev) dance. Collapse both into one helper:


const setContactCodeParam = useCallback(
  (code: string | null) => {
    setSearchParams(
      (prev) => {
        const next = new URLSearchParams(prev)
        if (code) next.set('contactCode', code)
        else next.delete('contactCode')
        return next
      },
      { replace: true }
    )
  },
  [setSearchParams]
)
Then handleSelectTenant becomes setContactCodeParam(tenant?.contactCode ?? null), and handleReset/the not-found effect call setContactCodeParam(null) instead of clearContactCodeParam.

albinrm (approved):

PR #669, opened 2026-07-08
Epic/mim 1838 kommunikationslogg

albinrm:


simonkropp (commented):
Some notes.. I could not find anything major but I haven't been able to test the flows.
Claude noted on some things, i excluded the most nitpicky ones. Worth looking into if they make sense.

**Correctness**

🔴 SMS delivery-report URL example points at the Keycloak-gated email webhook — services/communication/src/common/config.ts:16
The comment/example for INFOBIP__SMS_DELIVERY_REPORT_URL says https://<host>/webhooks/infobip. That's the email path (guarded by requireRole('infobip-webhook') in core/src/app.ts). The SMS path is /webhooks/infobip-sms, which lives on the public router and does its own ?token= check. Tele2/Infobip POST with only ?token= and no Keycloak bearer, so pointing them at /webhooks/infobip returns 401 and every SMS delivery report is dropped — recipient rows stay pending forever. core/.env.template:22 documents the correct path, so the two files contradict each other. Fix the comment/example to /webhooks/infobip-sms.

**Auto-dispatches in the parking-space flow (bilplatsflödet)**

🔴 The offer route reports a false failure on every offer — email.ts:206 returns 200, but communication-adapter/index.ts:135 checks !== 204. I audited all four routes: only the offer route is mismatched (accept / non-scored-approved / non-scored-denied all return 204 and match). So on every internal offer, create-offer.ts logs "Send Parking Space Offer Email … failed" and logger.errors — even though the email sent and the log row was written. This is the one that most needs fixing before merge: not just noise, it makes a genuine offer-email failure indistinguishable from the ever-present false one (alert fatigue / masked real failure). Fix = return 204 from the offer route (matching its siblings).

🟠 The external non-scored approved/denied flow ignores the adapter result entirely — external/index.ts:253 and :290 await the send but never check the returned AdapterResult.ok (unlike create-offer.ts and reply-to-offer.ts, which both check !result.ok). If the communication service is down or the send throws, the process still returns successful and emails the leasing team a "Godkänd/Nekad ansökan" summary — with no signal anywhere that the applicant was never notified. This is a genuine hole and worth handling, given these are exactly the hard-to-test auto-sends.

**app.ts + infobipSmsWebhookRoutes on the public router (no Keycloak)**

Looks fine to me but needs to be tested

🟠 The token is reflected back in the response body — including the 401. I confirmed this: generateRouteMetadata sets _links.self.href = ctx.href (generateRouteMetadata.ts:36), and ctx.href includes the full ?token=…. That metadata is spread into both the success body (:61) and the 401 body (:51). So the secret is echoed back to any caller — including an unauthorized prober who guessed wrong — and into anything capturing response bodies (APM, error trackers). This is the highest-value, easiest fix: strip the query string from the reflected href, or don't spread metadata on this route. Combined with the token also landing in ingress/proxy access logs (app.proxy = true) and being persisted on Tele2's side, the query-string token has a wide blast radius — but the response-body reflection is the one that's gratuitous and within your control.

albinrm:
**Correctness**

This is a comment only, fixed now though

**Auto-dispatches in the parking-space flow (bilplatsflödet)**

Both of these are preexisting. It does seem wrong, at the sanme time i dont reallly wanna touch it without ocnsulting an author (and I will need even more careful testing if i do touch them

**app.ts + infobipSmsWebhookRoutes on the public router (no Keycloak)**
Having the token in body is unnecessary exposutre, I removed metadata on that route

PR #674, opened 2026-07-16
Feature/mim 1897 mojlighet att schemalagga sms utskick

albinrm:


services/communication/src/services/communication-log-service/index.ts
```diff
@@ -86,6 +111,156 @@ export const routes = (router: OkapiRouter) => {
...
+          ctx.body = { dispatchId: ctx.params.id, cancelledRecipients: 0 }
+          return
+        }
+        if (!statuses.includes('scheduled')) {
+          ctx.status = 400
+          ctx.body = { error: 'NOT_SCHEDULED' }
+          return
+        }
+
+        await cancelScheduledBulk(dispatch.dispatch.channel, ctx.params.id)
```
Crewiz, line 170: Cancel isn't atomic: if Infobip cancel (158) succeeds but the DB update (159) throws, rows stay 'scheduled' for a bulk that won't send, and nothing else writes 'cancelled', the log shows "Schemalagt" permanently. Worth making a retry converge here.

services/communication/src/services/communication-log-service/adapters/db.ts
Crewiz: No status guard - a late delivery report can overwrite a 'cancelled' row with 'delivered'/'failed', contradicting the cancelledRecipients we returned. You can add .whereNot('status', 'cancelled') (only 'cancelled' needs protecting; the others are legitimately corrected by later reports).

apps/property-tree/src/shared/hooks/useBulkMessaging.ts
```diff
@@ -258,19 +269,27 @@ export function useBulkMessaging<TItem>({
...
           )
           .map((r) => ({ contactCode: r.id, phoneNumber: r.phone }))
 
-        const result = await sendBulkSms(recipients, message)
+        const result = await sendBulkSms(recipients, message, sendAt)
 
         toast({
-          title: 'SMS skickat',
-          description: `Skickades till ${result.content.totalSent} mottagare${
+          title: sendAt ? 'SMS schemalagt' : 'SMS skickat',
```
Crewiz: Toast picks its title from the request's sendAt, not from what the server did. The server treats a sendAt within ±30s of now as an immediate send (schedule.ts grace window) and returns no dispatchId. So if the user picks the earliest allowed minute and clicks in its second half, the message is sent now but the toast still says "SMS schemalagt", for something already gone and not cancellable

Key the title on result.content.dispatchId/scheduledFor instead. Same pattern in useSingleSms/useSingleEmail.

Crewiz (commented):
Some non-blocking CLAUDE.md convention cleanup:

Worth doing:
Axios in a new core adapter — dispatch-schedule.ts uses axios; new adapters should use openapi-fetch + createClient({ baseUrl }). Also leaves response shapes hand-typed instead of generated (drift risk).
New db fns lack try/catch + logger — deleteDispatch, setRecipientExternalIds, cancelScheduledRecipients, updateDispatchSendAt (db.ts) need logger.error({ err }, 'adapter.fnName'); today a DB failure throws raw with no adapter-named log.

Trivial:
catch (logError: any) (sms.ts:400, email.ts:736) — unknown works, rule forbids any.
'unknown error' codes (index.ts:175, 258) — return a specific code; also leaks raw error.message.
Email cap mis-documented as "90 days" in Swagger — it's 5.

Optional refactor:
The ~110-line schedule orchestration is near-duplicated across routes/sms.ts and routes/email.ts; a shared helper would stop the compensation logic drifting.

services/communication/src/services/communication-log-service/adapters/db.ts
albinrm: If a message we think is cancelled actually gets a delivery report, then we want to have that info. Will block bounced/failed for cancelled sends

Crewiz: That's better!

services/communication/src/services/communication-log-service/index.ts
albinrm: retry is probably smart

albinrm:
> Some non-blocking CLAUDE.md convention cleanup:
> 
> Worth doing: Axios in a new core adapter — dispatch-schedule.ts uses axios; new adapters should use openapi-fetch + createClient({ baseUrl }). Also leaves response shapes hand-typed instead of generated (drift risk). New db fns lack try/catch + logger — deleteDispatch, setRecipientExternalIds, cancelScheduledRecipients, updateDispatchSendAt (db.ts) need logger.error({ err }, 'adapter.fnName'); today a DB failure throws raw with no adapter-named log.
> 
> Trivial: catch (logError: any) (sms.ts:400, email.ts:736) — unknown works, rule forbids any. 'unknown error' codes (index.ts:175, 258) — return a specific code; also leaks raw error.message. Email cap mis-documented as "90 days" in Swagger — it's 5.
> 
> Optional refactor: The ~110-line schedule orchestration is near-duplicated across routes/sms.ts and routes/email.ts; a shared helper would stop the compensation logic drifting.

I dont want to deal with the generate step in core for another service, doesnt seem tio be a hard and fast rule since we arent doinfg that for our new contacts service i think

other feedback should be adressed

apps/property-tree/src/shared/hooks/useBulkMessaging.ts
albinrm: this was ostly a non-problem, (send within 30 sec would send immediately and then tell teh user it had been scheduled for 30 seconds away) however it keeps coming up in code agent reviews (my own as weel) so i adressed it to shut claude up

Crewiz (approved):

PR #677, opened 2026-07-27
feat: search invoices by number from top search

Crewiz:
Sök på fakturanummer, testas genom att söka på et (552607366096653 bör funka) och tar en till kundkortet + öppnar fliken med fakturan

### DIFF
--- a/apps/property-tree/src/features/search/hooks/useSearch.ts
+++ b/apps/property-tree/src/features/search/hooks/useSearch.ts
@@ -1,21 +1,32 @@
+import type { Invoice } from '@onecore/types'
 import { useQuery } from '@tanstack/react-query'
 
-import { searchService, workOrderService } from '@/services/api/core'
+import {
+  economyService,
+  searchService,
+  workOrderService,
+} from '@/services/api/core'
 import type { components } from '@/services/api/core/generated/api-types'
 import {
   type ContactSearchResult,
   tenantService,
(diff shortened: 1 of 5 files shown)


albinrm (commented):
Only one thing potentially blocking

Is the ID length of an invoice known? Because I think we opay credits per lookup to xledger, and if we call lookups on every phone number and personal number serach we do, that may be wasted money. if we know that ids are longer thatn say 10, or 12 digits we should probably limit it to that

Claude: Mobile deep link is a dead end. TenantTabsMobile ignores both tab and open ([TenantTabsMobile.tsx] has no nuqs usage), so selecting an invoice result on mobile lands on the accordion with nothing expanded. Pre-existing limitation of the tabs widget, not something you introduced — fine to defer, but worth a line in the PR.

Pre-existing, but now user-reachable: GraphQL string interpolation. [xledger-adapter.ts:913] interpolates the invoice number straight into the GraphQL query. Your frontend regex constrains what this UI sends, but the core route accepts any string from any authenticated client, and this PR is the first to wire a free-text input to it. Not this PR's bug — but a small escaping/validation fix in the economy service (or a digits-only check in the route) would be a cheap hardening follow-up.

Placeholder text: "Sök fastigheter, lägenheter, kunder..." ([CommandPalette.tsx:234 — add "fakturor" so anyone can discover the feature exists.

albinrm (approved):

PR #678, opened 2026-07-28
allow editing own listing comments (MIM-1688)

Crewiz:
## What & why

MIM-1688: adds the ability to **edit a posted comment** in the parking-space
listing "Notering/kommentar" log. Previously users could only create and
delete comments. Requested by Uthyrningen (edit own comments; not the event log).

## Changes

- New `PUT /comments/:targetType/thread/:targetId/:commentId` threaded through
  every tier: **leasing service → core → internal-portal backend → frontend**,
  mirroring the existing DELETE handler.
- **Ownership enforced in the backend** — only the comment's author can edit
  (401 `access-denied` otherwise); the "Redigera" button shows only on your own
  comments.
- New nullable `updatedAt` column (migration included) drives a subtle
  **"(redigerad)"** marker on edited comments.
- Inline edit UI in `CommentSection` (text + type), with local state kept in
  sync so an edit can't overwrite newer server content.

## Fixes made along the way (found while reviewing this change)

- **Auth bypass in DELETE:** the handler set `ctx.status = 401` for non-authors
  but was missing a `return`, so it deleted anyway. Now guarded (PUT enforces the
  same check).
- **Dead error mappers:** the comment error mappers used `declare enum` (no
  runtime object), so tailored messages never applied. Switched add/remove/update
  to real enums and surfaced the mapped messages in the add/edit/delete toasts.
- Removed an unreachable 404 branch in the core update adapter.

## Notes / out of scope

- `createdAt`/`updatedAt` are typed `z.date()` but cross services as JSON strings;
  latent only (nothing re-parses transported comments today) — left as-is.

### DIFF
--- a/apps/internal-portal/backend/src/services/leasing-service/comments.ts
+++ b/apps/internal-portal/backend/src/services/leasing-service/comments.ts
@@ -92,6 +92,7 @@ export const routes = (router: KoaRouter) => {
       } else if (comment.authorId !== ctx.session?.account.username) {
         ctx.status = 401
         ctx.body = { error: 'access-denied', ...metadata }
+        return
       }
 
       const result = await coreAdapter.removeComment(threadId, commentId)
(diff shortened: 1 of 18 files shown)


core/src/adapters/leasing-adapter/comments.ts
```diff
@@ -68,4 +68,29 @@ const removeComment = async (
...
+
+  if (response.status === 200) {
+    return {
+      ok: true,
+      data: response.data.content,
+      statusCode: response.status,
+    }
+  }
+
+  return { ok: false, err: 'unknown', statusCode: response.status }
```
albinrm: Stock axios has a default validateStatus of "2xx only": any response outside 2xx makes the promise reject (throw an AxiosError) instead of resolving with a response object.

const response = await axios.put(...)        // ← throws on 404, never returns
if (response.status === 200) { return { ok: true, ... } }
return { ok: false, err: 'unknown', statusCode: response.status }  // ← dead code

Where the thrown error lands. The core route handler has no try/catch around leasingAdapter.updateComment(...), so the AxiosError bubbles up to Koa's generic error handling and core responds 500 — not the 404 documented 

In the internal-portal backend's core-adapter there is a 404 mapping:

if (err instanceof AxiosError && err.response?.status === 404) {
  return { ok: false, err: 'not-found', statusCode: 404 }
}

apps/internal-portal/backend/src/services/leasing-service/comments.ts
```diff
@@ -106,4 +107,64 @@ export const routes = (router: KoaRouter) => {
...
+      )
+      if (!comment) {
+        ctx.status = 404
+        ctx.body = { error: 'not-found', ...metadata }
+        return
+      } else if (comment.authorId !== ctx.session?.account.username) {
+        ctx.status = 401
+        ctx.body = { error: 'access-denied', ...metadata }
+        return
+      }
```
albinrm: 401 vs 403 — access-denied returns 401, which means "unauthenticated"; 403 is the semantically correct code. Consistent with the delete route, so only worth changing if you touch both.

albinrm (commented):
Some Minor points ive made at the specific lines of code, really non blocking but preffered to resolve

Some  additional claude nits :

Core PUT route skips body validation ([core comments.ts:140] a bare <UpdateCommentRequest> cast, while both leasing and the portal backend safeParse. It matches the existing POST pattern, so fine to leave, but core is the public orchestration API and the schema is already imported; a safeParse is three lines.

Duplication in the portal backend — the PUT route repeats the delete route's fetch-thread → find-comment → check-ownership block verbatim. A small shared helper would keep the authorization logic in one place, which matters more now that it is the authorization layer.

Swagger requestBody is type: object with no properties, so the generated api-types.ts request body is untyped. Listing type/comment (or referencing the schema) would make the generated client types useful. Same nit applies to the existing routes.

Silent no-op on empty save — if (!editComment.trim()) return in handleSave; disabled={!editComment.trim()} on the Spara button would communicate better.

albinrm (approved):
everything ive mentioned seems to be adressed. I ahvent spun up the portal and tested editing a comment but assuming you have done so its ready to merge.

PR #683, opened 2026-07-29
MIM-1825 - Flera filer i ströfaktura

Crewiz:
## MIM-1825 — Flera filer i ströfaktura

Users can now attach multiple files to a miscellaneous invoice (ströfaktura).
Since Xledger only accepts a single file per invoice base item (see Lina's
comment on the card), the files are merged into one PDF client-side before
submit. The entire backend chain
(core → economy → Xledger) is untouched and still receives exactly one file.

### How it works

- New helper `mergeFilesToPdf.ts` (property-tree): merges PDF/JPG/PNG into a
  single `strofaktura-bilagor.pdf` using `pdf-lib` (same version already used
  in keys-portal and services/keys). Dynamically imported so it stays out of
  the main bundle. Images are scaled down to fit A4 (never upscaled).
- `AdditionalInfoSection`: multi-file picker, one row per file with remove,
  dedupe by name+size, inline validation messages.
- `MiscellaneousInvoiceForm`: 0 files → no attachment; 1 file → sent as-is,
  untouched (today's behavior, any file type); ≥2 files → merged before
  submit. If a file can't be read (encrypted/corrupt PDF), the submit is
  aborted with a toast naming the file — never silently partial.

### Behavior change to be aware of

Multi-file selections are restricted to **PDF/JPG/PNG** — forced by
mergeability. A *single* attachment of any type still works exactly as
before. The invariant: the attachment list is either one file of any type,
or several mergeable files.

### Testing

- [x] Multiple files merged into one PDF and submitted — invoice base item
      created in Xledger demo (200 end-to-end, attachment referenced in the
      GraphQL mutation)
- [x] Visual check of the merged PDF inside Xledger demo, opened the invoice
      base item (2026-08-03) and verified the attachment: all pages present
      and the image page scaled to A4
- [x] Typecheck (repo root) and lint clean

### DIFF
--- a/apps/property-tree/package.json
+++ b/apps/property-tree/package.json
@@ -22,10 +22,10 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@dnd-kit/core": "^6.3.1",
     "@deck.gl/core": "^9.0.38",
     "@deck.gl/layers": "^9.0.38",
     "@deck.gl/react": "^9.0.38",
+    "@dnd-kit/core": "^6.3.1",
     "@headlessui/react": "^1.7.18",
     "@onecore/types": "workspace:*",
     "@radix-ui/react-accordion": "^1.2.12",
@@ -59,6 +59,7 @@
     "lucide-react": "^0.462.0",
     "maplibre-gl": "5.7.0",
     "nuqs": "^2.8.6",
+    "pdf-lib": "^1.17.1",
     "react": "^18.3.1",
     "react-day-picker": "^9.13.0",
     "react-dom": "^18.3.1",
--- a/apps/property-tree/src/features/economy/components/AdditionalInfoSection.tsx
+++ b/apps/property-tree/src/features/economy/components/AdditionalInfoSection.tsx
@@ -26,7 +26,7 @@ interface AdditionalInfoSectionProps {
   attachedFiles: File[]
 }
 
-const fileKey = (file: File) => `${file.name}:${file.size}`
+const fileKey = (file: File) => `${file.name}:${file.size}:${file.lastModified}`
 
 export function AdditionalInfoSection({
   selectedProject,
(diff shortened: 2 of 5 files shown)


albinrm (changes requested):
My reading of it is it looks fine. Ran thorugh the revfiew skill in claude and it came up with the following

out of thesxe i think opnly 1 is  potentially blocking

1. The accept attribute contradicts the single-file-any-type design — 
The code deliberately preserves the old behavior of attaching a single file of any type ("like before"), and there's even a dedicated error path for when a non-mergeable file is already attached. But the new accept=".pdf,.jpg,.jpeg,.png,..." filters the picker to those three types always, so attaching that single .docx — which the logic supports — now requires the user to switch the picker to "All files". Previously the input had no accept at all, so this is a quiet regression for anyone who attaches Word/Excel underlag today. Two consistent options: make accept conditional (unrestricted when attachedFiles.length === 0, restricted otherwise), or decide that only PDF/JPG/PNG are supported at all and delete the single-any-type branch plus its error path. Either is fine; the current middle ground is the only combination I'd push back on.

2. EXIF orientation is ignored when embedding photos 
pdf-lib's embedJpg does not apply EXIF rotation. Portrait photos from phones are often stored landscape with an orientation flag, so they'll render rotated 90° in the merged PDF. Since "photos of underlag straight off a phone" is presumably the main multi-file use case, test this with a real portrait phone photo before merging. If it reproduces, the standard fix is to normalize images through a canvas (createImageBitmap + canvas re-encode applies orientation) before embedding.

3. fileKey can silently drop distinct files — 

name:size treats two different files with the same name and size as duplicates — e.g. kvitto.jpg from two different folders — and the second is dropped without any message. Adding file.lastModified to the key makes collisions near-impossible and costs nothing. Minor.

-- i find this to be incredibly unlikely (bite size identical and naem identical files, itll be a one in 10k occurance given two identical names)

Nits 
4 save() already returns a Uint8Array; the new Uint8Array(mergedBytes) copy is redundant unless it's there to appease BlobPart typing (if it compiles without, drop it).
5 MergeFileError.fileName is never read (the message already embeds it) — harmless, but could go.

albinrm (approved):
One thing that could be done is making the resize only occur on a branch where its needed. normalizeJpeg round-trips all JPEGs through canvas at quality 0.92 to fix EXIF rotation, but most scanned/exported JPEGs have no rotation, so reduces quality a little.

Non blocking.

Claude produced this comment:

Unrelated lockfile churn — [pnpm-lock.yaml] contains a jest/ts-jest peer-resolution reshuffle (@types/node@24 → @types/node@20 + ts-node variant) in one service alongside the pdf-lib addition. Harmless, but it's noise in the PR — a fresh pnpm install from a clean checkout might drop it.

PR #688, opened 2026-08-03
Feature/mim 1911 kommunikationslogg oversiktsvy

albinrm:


services/communication/src/services/communication-log-service/dispatch-status.ts
```diff
@@ -0,0 +1,42 @@
...
+): DispatchStatus {
+  const has = (s: RecipientStatus) => (counts[s] ?? 0) > 0
+  const total = Object.values(counts).reduce<number>((a, b) => a + (b ?? 0), 0)
+
+  // Empty is a defensive default only — real dispatches always have >=1
+  // recipient. The SQL status predicates in db.ts match nothing for an empty
+  // dispatch, so a zero-recipient row would badge 'failed' but not appear under
+  // the failed filter (out of domain in practice).
+  if (total === 0) return 'failed'
+  if (has('scheduled')) return 'scheduled'
```
Crewiz: "Scheduled" wins the derivation even after the bulk fires, so a fired bulk shows "Schemalagt" until every terminal delivery report lands (forever if one is lost), and it's excluded from status=delivered filters. Maybe we could do a sendAt < now tiebreak.

services/communication/src/services/communication-log-service/adapters/db.ts
```diff
@@ -307,3 +328,244 @@ export async function getCustomerMessages(
...
+  ctx: Context
+): Promise<PaginatedResponse<communication.DispatchListItem>> {
+  const query = db<Dispatch>('dispatch').select('dispatch.*')
+
+  if (params.channel?.length) query.whereIn('dispatch.channel', params.channel)
+  if (params.messageType?.length)
+    query.whereIn('dispatch.messageType', params.messageType)
+  if (params.triggeredByUser)
+    query.where('dispatch.triggeredByUser', params.triggeredByUser)
+  if (params.source === 'automatic')
```
Crewiz: Dispatches with NULL triggeredByUser fall through both source filters, they're not the automatic literal, but the manual branch does whereNotNull, so they match neither. The work-order SMS/email paths log NULL today (only parking-space email has the fallback), so those sends are only findable with no source filter at all. Maybe treat NULL as automatic here, or add the fallback in the work-order log helpers?

services/communication/migrations/202607270000_create_dispatch_audience_criterion.js
Crewiz, line 1: Do we care about dispatches logged before this ships? They have audienceCriteria JSON but no criterion rows, so they'll show an empty Målgrupp and won't match audience filters. If yes, a one-time backfill should be cheap since audienceCriteriaToRows already matches the stored shape.

services/communication/src/services/communication-log-service/adapters/db.ts
```diff
@@ -307,3 +328,244 @@ export async function getCustomerMessages(
...
+}
+
+/**
+ * Paginated dispatch search. Filters on the dispatch row directly, on recipient
+ * rows via EXISTS (contactCode, derived status), and on audience criteria via
+ * EXISTS over dispatch_audience_criterion. Each returned dispatch is enriched
+ * with a per-status recipient rollup, its derived status, and its audience
+ * criteria. See MIM-1911.
+ */
+export async function searchDispatches(
```
Crewiz, line 394: searchDispatches and getDispatchRecipients (line 548) are missing the try/catch + logger.error pattern the rest of this file uses, so errors from these won't show up in the structured logs.

apps/property-tree/src/entities/dispatch/hooks/useDispatchFilters.ts
```diff
@@ -0,0 +1,135 @@
...
+    rawSortOrder === 'asc' || rawSortOrder === 'desc' ? rawSortOrder : undefined
+
+  const searchParams = useMemo<DispatchSearchQueryParams>(
+    () => ({
+      q: filters.debouncedSearch || undefined,
+      channel: selectedChannels.length > 0 ? selectedChannels : undefined,
+      status: selectedStatuses.length > 0 ? selectedStatuses : undefined,
+      messageType:
+        selectedMessageTypes.length > 0 ? selectedMessageTypes : undefined,
+      source: (source as 'manual' | 'automatic') || undefined,
```
Crewiz, line 61: source gets cast and forwarded as-is, so a stale bookmark like ?source=all becomes a 400 and blanks the table, unlike sortBy/sortOrder above which are whitelisted and silently dropped when invalid. Same applies to channel/status. Could we validate these the same way?

apps/property-tree/src/entities/dispatch/hooks/useDispatchFilters.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export function useDispatchFilters() {
+  const filters = useUrlFilters({
+    filterKeys: FILTER_KEYS,
+    pageSize: PAGE_SIZE,
+  })
+
+  const { searchParams: urlSearchParams } = filters
+
+  const selectedChannels = useMemo(
```
Crewiz, line 30: The next-page prefetch re-runs on every params change, so each debounced keystroke that yields >1 page also fires a full search for page 2 that mostly never gets visited. Maybe only prefetch on actual pagination, since keepPreviousData already smooths out the filtering case?

apps/property-tree/src/pages/DispatchCentrePage.tsx
```diff
@@ -0,0 +1,271 @@
...
+  non_scored_parking_space_denied: 'Bilplats nekad',
+}
+const categoryLabel = (messageType: string) =>
+  CATEGORY_LABELS[messageType] ?? messageType
+
+const categoryOptions = Object.entries(CATEGORY_LABELS).map(
+  ([value, label]) => ({ value, label })
+)
+
+const formatTimestamp = (iso: string): string =>
```
Crewiz: This is identical to formatScheduleTimestamp in shared/lib/schedule.ts (and TenantCommunicationTabContent has a third copy), should the shared one be used instead?

Crewiz (changes requested):

services/communication/migrations/202607270000_create_dispatch_audience_criterion.js
albinrm: DIspatches logged before do not use audienceCriteria at all, it not currently being supplied, and it cannot really be backfilled.

services/communication/src/services/communication-log-service/adapters/db.ts
albinrm: Odoo does supply triggeredByUser but i suppose it wouldnt hurt to put NULL into manual. Automatic sends are completely under our control considering they have their own routes, so we can alwasy make sure its not null there.

apps/property-tree/src/entities/dispatch/hooks/useDispatchFilters.ts
albinrm: I think i prefer just removing the whitelist logic for sortBy/sortOrder. Requests shouldnt silently drop filters, because they will behave differently than users expect.

apps/property-tree/src/entities/dispatch/hooks/useDispatchFilters.ts
albinrm: Adding a delay on parameter change -> fethc next page, to counteract this. We should maybe look at implementign the same in other search pages.

services/communication/src/services/communication-log-service/adapters/db.ts
albinrm: yea adding

apps/property-tree/src/pages/DispatchCentrePage.tsx
albinrm: Adressed

services/communication/src/services/communication-log-service/dispatch-status.ts
albinrm: Adressed

Crewiz (approved):
LGTM!

PR #690, opened 2026-08-06
Fix:MIM-1888 Correct Acquisition Date in PropertyDetailsSchema

mittistormen:
`acquisitionDate` was typed as `z.string()` in `PropertyDetailsSchema` but
  Prisma returns it as a `Date` object, causing a parse error on GET /properties/:code.

  Changed to `z.date()` to match the actual runtime type, consistent with how
  `registrationDate` is already typed in the same schema.

### DIFF
--- a/services/property/src/types/property.ts
+++ b/services/property/src/types/property.ts
@@ -70,7 +70,7 @@ export const PropertyDetailsSchema = z.object({
   consolidationNumber: z.string().nullable(),
   ownershipType: z.string(),
   registrationDate: z.date().nullable(),
-  acquisitionDate: z.string().nullable(),
+  acquisitionDate: z.date().nullable(),
   isLeasehold: z.number().int(),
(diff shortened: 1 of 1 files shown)


albinrm (approved):
checked that this 

`[10:29 AM]{
  "reason": "[\n  {\n    \"code\": \"invalid_type\",\n    \"expected\": \"string\",\n    \"received\": \"date\",\n    \"path\": [\n      \"acquisitionDate\"\n    ],\n    \"message\": \"Expected string, received date\"\n  }\n]",`

is the bugs error, and this 

`SELECT COLUMN_NAME, DATA_TYPE, IS_NULLABLE
FROM INFORMATION_SCHEMA.COLUMNS
WHERE TABLE_NAME = 'bafst'

----------------------------

COLUMN_NAME	DATA_TYPE	IS_NULLABLE
fangesdat	date	YES`

is true in the db

PR #691, opened 2026-08-07
fix(leasing): return 404 instead of 500 when contact is not a tenant …

Crewiz:
### Problem

Searching for a contact in property-tree calls core's `GET /tenants/by-contact-code`, which in turn calls leasing's `GET /contacts/:contactCode/tenant`. When the contact isn't a tenant (e.g. P089052), leasing responded **500**, which propagated as ERROR logs in both leasing and core on every normal search — even though "contact is not a tenant" is an expected outcome.

### Changes

- **leasing**: the tenant endpoint now responds **404** for the expected not-a-tenant outcomes (`contact-not-tenant`, `no-valid-housing-contract`, `contact-leases-not-found`), with the semantic code in the body's `type` field. Genuine failures still respond 500. These outcomes are now logged as `info` instead of `error`.
- **core**: `leasing-adapter.getTenantByContactCode` maps the 404 body `type` to its existing error union without error-logging, and `/tenants/by-contact-code` propagates 404 (previously re-mapped to 500). Only the known leasing error types are accepted on 404 — an unrecognized 404 (e.g. a proxy/gateway 404 or misrouted service URL) is logged as an error and returned as `unknown`, so an infrastructure fault can't masquerade as "not a tenant". The now-obsolete `contact-leases-not-found` mapping in the catch path was removed (during deploy skew that case simply falls through to the pre-existing 500 behavior).
- **internal-portal (backend)**: the core adapter reads the error `type` from 404 responses, in addition to the existing 500 handling (kept for compatibility with core versions that predate this change). The `validate-tentant` route also got an explicit `contact-not-found` branch (403 with a proper body) — previously that case was masked as `contact-not-tenant` by core's adapter; without the new branch it would have surfaced as a bodyless 500 with client retries.

### Consumers verified

All consumers of the endpoint chain were mapped and verified:

- **Parking-space rental flow** (internal-portal `validate-tentant`): distinguishes non-tenants by error code and blocks with 403 — the codes are preserved end-to-end through the 404 body, so the flow behaves exactly as before. Verified this flow does **not** rely on the error/500 to permit or deny anything it shouldn't.
- **Core `/listings` filtering**: already treats these codes as expected soft-fails — unaffected. Note: `contact-not-found` now reaches this code path under its true name (it was previously collapsed into `contact-not-tenant`) and is error-logged there by existing design — kept as-is since that log is intentional for authenticated contacts.
- **Work-order creation**: treats any adapter error as 500 regardless of code — unaffected.
- **Property-tree contact search**: falls back to `/contacts/:contactCode` on any error — same behavior, without the log spam.

### Follow-up (not in this PR)

The tenant error codes are string literals maintained in leasing, core and internal-portal separately. A shared `GetTenantErrorCodes` in `libs/types/src/error-types` (following the existing `CreateOfferErrorCodes` pattern) would make the mapping compile-time-checked across packages.

### DIFF
--- a/apps/internal-portal/backend/src/services/leasing-service/adapters/core-adapter.ts
+++ b/apps/internal-portal/backend/src/services/leasing-service/adapters/core-adapter.ts
@@ -150,6 +150,15 @@ const getTenantByContactCode = async (
 
     return { ok: true, data: result.content }
   } catch (err) {
+    if (err instanceof AxiosError && err.response?.status === 404) {
+      // Expected when the contact is not a tenant. The body `type` field
+      // distinguishes contact-not-found, contact-not-tenant and
+      // no-valid-housing-contract.
+      return { ok: false, err: err.response?.data?.type, statusCode: 404 }
+    }
+
+    // Core versions before MIM-1940 responded 500 for not-a-tenant outcomes,
+    // with the same `type` field in the body.
     if (err instanceof AxiosError && err.response?.status === 500) {
       return { ok: false, err: err.response?.data?.type, statusCode: 500 }
     }
--- a/apps/internal-portal/backend/src/services/leasing-service/adapters/core-adapter.ts
+++ b/apps/internal-portal/backend/src/services/leasing-service/adapters/core-adapter.ts
@@ -150,11 +150,15 @@ const getTenantByContactCode = async (
 
     return { ok: true, data: result.content }
   } catch (err) {
-    if (err instanceof AxiosError && err.response?.status === 404) {
-      // Expected when the contact is not a tenant. The body `type` field
-      // distinguishes contact-not-found, contact-not-tenant and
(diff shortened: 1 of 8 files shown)


mittistormen (changes requested):
Förstår inte varför du tar bort type ur err här? Borde inte det vara mer specifikt/tydligt att skicka tillbaka det när det finns?

Crewiz:
> Förstår inte varför du tar bort type ur err här? Borde inte det vara mer specifikt/tydligt att skicka tillbaka det när det finns?

Bra poäng! Det spelar ingen roll i praktiken just nu (enda anroparen kollar bara koderna som kommer via 404), men det blir tydligare så. Ändrade till att alltid returnera type när den finns, oavsett statuskod 👍

mittistormen (approved):
Snyggt!!

